PR #5, opened 2022-06-12
Added ContentPage along with its template

Hitansh-Shah:
I added a new app `content` and defined a new page `ContentPage` along with its template inside it. I added some restrictions like `max_count` and `subpage_types` to both `ContentPage` and `HomePage`. @allcaps


allcaps (changes requested):
- The `*.pyc` files should not be in version control. Add them to the .gitignore
- Files that we don't use should be removed. We don't use views.py and test.py
- We've got two migration files. The rename is unnecessary and can be in the initial migration. Please recreate the migrations (drop your db and the migration files and run makemigrations)
- Nitpick, but the current commit message is `Added ContentPage along with its template`. We should write all commit messages in imperative mode. Like you tell the project what to do: "Add ContentPage and template". You can change a commit message (assuming this is the last commit) with: `git commit --amend`

Hitansh-Shah:
I have done the changes you requested. Please take a look @allcaps

allcaps (changes requested):

allcaps (approved):

PR #6, opened 2022-07-01
Added tests

Hitansh-Shah:
@allcaps I added some tests and also installed django-extensions on this branch. There is also one commit for adding fixtures. However the tests are not using fixtures currently and we can delete/revert that commit later if we want. Also this branch was branched from the branch `contentpage` (the one which is open for merging).


content/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
allcaps, line 1: We don't use this file. So it can be deleted.

content/blocks.py
```diff
@@ -0,0 +1,5 @@
+from wagtail.core.blocks import RichTextBlock
+
+class ContentBlock(RichTextBlock):
+    class Meta:
+        template = 'content/content_block.html'
```
allcaps, line 5: We don't enforce code style yet, but is it a good practice to end files with a new line.

home/fixtures/test.json
```diff
@@ -0,0 +1,122 @@
+[
```
allcaps, line 1: Why do we have this file? Is it used in test?

These are a pain to maintain. Better create content programatically.

guide/tests/test_page_response.py
```diff
@@ -0,0 +1,58 @@
+from django.test import TestCase
+from wagtail.models import Page
+from content.models import ContentPage
+from home.models import HomePage
+from wagtail.models.sites import Site
+class TestPageResponseStatusCode(TestCase):
+    # fixtures = ["test.json"]
```
allcaps, line 7: Remove comment

guide/settings/production.py
```diff
@@ -1,7 +1,7 @@
 from .base import *
 
 DEBUG = False
-
+STATICFILES_STORAGE = "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
```
allcaps, line 4: Can we keep this in base? Just override `STATICFILES_STORAGE` in dev or test settings if you need something else. 

At least keep it with the comment.

home/fixtures/test.json
Hitansh-Shah: This is not used currently. I created this a while ago when I was trying out fixtures. I didn't delete it because I thought we may need it in future. I'll delete it.

guide/settings/production.py
Hitansh-Shah: So I was facing some issues while running tests and I asked on support and Matt suggested to comment this out in base and re-add it in production. Here is the link to the [discussion](https://wagtailcms.slack.com/archives/C81FGJR2S/p1655212725331549)

PR #8, opened 2022-07-02
Rework project structure

allcaps:


Hitansh-Shah:
Cool! Shouldn't we also add `black`, `isort` and `flake8` in `requirements/development.txt`? Also is it safe to delete the old folders for all the apps in the root directory? They only seem to contain `pycache`

allcaps:
Yes, you are right. We need to add those requirements! 👍 Updated.

You can safely delete those directories. Git doesn't do that for you.

Makefile
```diff
@@ -1,7 +1,16 @@
...
+	black --check --diff apps
+	python manage.py test
+
 format:
-	black --target-version py37 .
-	isort .
+	isort apps
+	black apps
 
 lint:
```
Hitansh-Shah, line 14: Wagtail also runs `flake8` while linting. Should we add it?

setup.cfg
```diff
@@ -0,0 +1,12 @@
+[flake8]
+max-line-length = 88
+max-complexity = 6
+exclude = */migrations/*
+
+[tool:isort]
```
Hitansh-Shah, line 6: Shouldn't it be just `[isort]` instead of `[tool:isort]`? According to isort config guide it should be the former one.

apps/content/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class ContentConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "apps.content"
```
Hitansh-Shah, line 6: Just had a doubt. Is the `name` property here affect the imports? I mean if we change the name to something else say just `content`, would we then write `from content.models ...` instead of `from apps.content.models ...` ?

Hitansh-Shah:
@allcaps, fixtures file `test.json` is still there because I forgot to delete it and the PR got merged.

apps/content/apps.py
allcaps: > Full Python path to the application, e.g. 'django.contrib.admin'.
https://docs.djangoproject.com/en/4.0/ref/applications/#django.apps.AppConfig.name

`apps/__init__.py` makes `apps` directory a python module. Therefore you observed the changes correctly. We do need to import `from apps.content.models ...` however, the `AppConfig.name` just needs to reflect the path. It doesn't dictate the path.

Also note that other apps don't have this config. If we would write future proof code, we might add it. We also need to update the `INSTALLED_APPS`  values. But that is for another PR.

Hitansh-Shah (approved):

apps/content/apps.py
Hitansh-Shah: Thanks for the explanation. I understand it now :)

PR #9, opened 2022-07-03
Add frontend

allcaps:
This PR depends on the work in https://github.com/wagtail/guide/pull/8 , that PR needs to be merged first.


Hitansh-Shah (approved):
Hey @allcaps, so if I am not wrong all we have done in this setup is create a config file for `webpack` which takes `frontend/static_src/js/main.js` and `frontend/static_src/css/main.css` as entry point and outputs in `frontend/static` directory and also creates a `manifest.json` file because we used `django-manifest-loader`. Also I guess we have also installed `bootstrap` and imported components in `frontend/static_src/scss/main.scss`. I assume all the stuff done in settings file is related to the configuration of this setup.

Also I would need to run 2 servers now, one by `yarn start` (in dev) and other `./manage.py startserver`, right?

allcaps:
Your first paragraph is completely correct.

During deployment, we run `yarn build` to generate `/frontend/static/` containing all frontend assets (css, js, fonts, images).
We will push this frontend build to the server. Since `manifest.json` is updated, the production server will use the new assets. Note that these filenames contain a hash, thus invalidating the browser cache.

The server (NGINX or whatever) will serve the files form `static` directory.

`yarn start` is for development, a file watcher will rebuild the frontend each file change.

Hitansh-Shah:
Cool! Thanks 🙌

PR #26, opened 2022-07-22
Add main nav

allcaps:
# Screenshot of work done

Screenshot 2022-07-23 at 21 11 15

# Issue

I raised this in our Slack channel. Added it here as archive. 

I tried to create the navigation and am fighting the data structure a bit.
- It is impossible to navigate from 'Tutorials' to 'How-to'. The main navigation is missing.
- The collapsed navigation hide relevant information.
- The collapsed navigation is harder to use. As a user needs to click, click, click to expand the collapsed sections.

In the current guide, the page 'Finding your way around' duplicates the navigation. It is redundant.
https://docs.wagtail.org/en/stable/editor_manual/finding_your_way_around/index.html

If we drop the redundant 'in between pages', we loose a level, we can drop the collapse feature :tada:

Mockup:

nav

1. The current design.
2. Drop the lowest level, the contents of these pages can be on "Finding your way around".
3. OR, drop "Finding your way around", the underlying pages are lifted to a higher level.
4. Add the section titles. No need for a `+` icon. A click on the item can trigger a page load.

Note, that I couldn't adjust the boldness of the text in Photoshop. "The Explorer menu, ..." should have the same weight as "Introduction, Getting started, ..."

The proposed navigation is simpler. Easier to understand, easier to navigate, and easier to make and maintain.

### DIFF
--- a/apps/core/management/commands/buildfixtures.py
+++ b/apps/core/management/commands/buildfixtures.py
@@ -4,11 +4,14 @@ import uuid
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.template.defaultfilters import slugify
+from faker import Faker
 from wagtail.documents import get_document_model
 from wagtail.images import get_image_model
 from wagtail.models import Collection, Locale, Page
 
 from apps.core.factories import ContentPageFactory, HomePageFactory
+
+fake = Faker()
 
 
 class Command(BaseCommand):
@@ -39,9 +42,28 @@ class Command(BaseCommand):
         for title, subpage_titles in [
             [
                 "Tutorial",
-                ["Introduction", "Getting started", "Finding your way around"],
+                [
+                    "Getting started",
+                    "Dashboard",
+                    "Images",
+                    "Documents",
+                    "Page explorer",
+                    "Page editor",
+                    "Search",
+                ],
             ],
-            ["How-to", ["Add a user", "Assign permissions" "Tag content"]],
+            [
+                "How-to",
+                [
+                    "Add a user",
+                    "Assign permissions",
+                    "Create rich text",
+                    "Work with collections",
+                    "Tag content",
+                    "Add a focus area to an image",
+                    "Order pages",
+                ],
+            ],
             [
                 "Explanation",
                 [
@@ -53,8 +75,9 @@ class Command(BaseCommand):
             [
                 "Reference",
                 [
-                    "Dashboard",
-                    "Page editor",
+                    "Glossary",
+                    "Rich text shortcuts",
+                    "Groups and permissions",
                     "Workflows",
                 ],
             ],
@@ -70,7 +93,6 @@ class Command(BaseCommand):
                     title=subpage_title,
                     slug=slugify(subpage_title),
                 )
-
             self.home.sections = json.dumps(
                 [
                     {
--- a/apps/core/models/content.py
+++ b/apps/core/models/content.py
@@ -6,6 +6,7 @@ from ..blocks import CONTENT_BLOCKS
 
 
 class ContentPage(Page):
+    show_in_menus_default = True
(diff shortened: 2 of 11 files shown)


apps/core/templates/core/content_page.html
```diff
@@ -1,11 +1,17 @@
 {% extends "base.html" %}
 
-{% load static %}
-{% load wagtailcore_tags %}
+{% load static wagtailcore_tags %}
```
Hitansh-Shah, line 4: Why haven't we loaded our custom tag?

apps/core/templatetags/core_tags.py
```diff
@@ -0,0 +1,36 @@
...
+            info["close"] = list(range(0, prev_depth - depth))
+        info = {"open": open_, "close": [], "level": depth - start_depth}
+        annotated_list.append((page, info))
+        prev_depth = depth
+    if start_depth and start_depth > 0:
+        info["close"] = list(range(0, prev_depth - start_depth + 1))
+    return {
+        "current_page": context.get("page"),
+        "annotated_list": annotated_list,
+    }
```
Hitansh-Shah, line 36: Can you brief what this tag does? I found it quite complex to understand just by reading.

Hitansh-Shah (commented):
Hey @allcaps , I tried to understand all the changes and additions, I have few doubts and I have commented them. 
Also I feel the code is becoming more and more complex with every PR. I may ask even more questions now as I work on the code.

apps/core/management/commands/buildfixtures.py
```diff
@@ -65,12 +68,19 @@ def create_content_pages(self):
...
                 )
-
+                if subpage_title == "Finding your way around":
+                    for subsubpage_title in [
+                        "The Dashboard",
+                        "The Explorer menu",
+                        "Using search",
+                        "The Explorer page",
+                    ]:
+                        ContentPageFactory(parent=subpage, title=subsubpage_title)
```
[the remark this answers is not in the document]
allcaps: You understood correctly, I've added those pages to see the effect. However, it is not realistic fixture data, so I updated it.

apps/core/models/content.py
```diff
@@ -6,6 +6,7 @@
 
 
 class ContentPage(Page):
+    show_in_menus_default = True
```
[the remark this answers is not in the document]
allcaps: Because we want all content pages in menus. We want to opt-out instead of opt-in. As the majority of pages should be in the navigation.

apps/core/templates/base.html
```diff
@@ -30,7 +30,9 @@
         {% include "components/header.html" %}
 
         <div class="container-lg px-md-5">
+          <div class="row">
           {% block content %}{% endblock %}
+          </div>
```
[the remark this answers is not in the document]
allcaps: The Bootstrap row activates the columns. Without, the columns would be stacked.

See https://jsfiddle.net/allcaps/Lbj1xqd9/1/

apps/core/templates/core/content_page.html
allcaps: Good question!

In `core_tags.py` we have:
```
from django.template.defaultfilters import register
```

This makes `@register.inclusion_tag` register as defaultfilter. But that should be:
```
from django import template

register = template.Library()
```

I'll update the pr.

apps/core/templatetags/core_tags.py
allcaps: I'll add a comment, split this out, and I'll add a test.

It does this https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.get_annotated_list

But on a filtered queryset (live and in menu).

Hitansh-Shah:
Thanks @allcaps . I think we can merge this PR.

PR #33, opened 2022-08-01
add favicon

Hitansh-Shah:


laymonage (approved):
Nice! This looks good to me, but we might want to confirm which favicon we want to use (i.e. the one on https://wagtail.org vs https://docs.wagtail.org) before merging.

vossisboss:
@Hitansh-Shah @laymonage Let's use the one from https://docs.wagtail.org/ for now and merge this PR. If I have a spare moment with Ben, I'll ask him if we should go for something more unique.

PR #34, opened 2022-08-02
Add meta link to english version of pages

Hitansh-Shah:


apps/core/templates/base.html
```diff
@@ -16,6 +16,11 @@
         <meta name="description" content="" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
 
+        {# canonical url #}
+        {% if page.locale.language_code and not page.locale.language_code == "en-latest" %}
+            <link rel="canonical" href="{{ page.full_url|canonical_url:page.locale.language_code }}" >
```
laymonage, line 21: I'm not very familiar with SEO, but it seems Google recommends the use of `<link rel="alternate"` for multilanguage sites: https://developers.google.com/search/docs/advanced/crawling/localized-versions

and they also no longer recommend using `<link rel="canonical"`: https://developers.google.com/search/blog/2011/12/new-markup-for-multilingual-content#annotating-pages-as-substantially-similar-content

This page that used to recommend `<link rel="canonical"` now has a warning that points to the "general guidelines for multilingual and multi-regional content" page, (the first link).

Again, I'm not very familiar with SEO, so I might be wrong!

vossisboss: @laymonage @Hitansh-Shah Let me run this one of our seo folks at Torchbox and see what they recommend. I'll report back here with my findings.

thibaudcolas: Canonical isn’t meant for this, unless we wanted all languages other than english to not be indexed (which is silly). `rel="alternate"` with a `hreflang` is definitely the way to go.

Something like:

```jinja
<link rel="alternate" hreflang="en" href="{{ page.full_url|canonical_url:page.locale.language_code }}" >
```

This will only create a link from each translated version to english. Normally you’d want each version to link to all other versions, however for the time being the above should be a good start.

apps/core/templatetags/core_tags.py
```diff
@@ -20,3 +20,7 @@ def navigation(context):
         "current_page": context.get("page"),
         "annotated_list": Page.get_annotated_list_qs(pages),
     }
+
+@register.filter
+def canonical_url(value, arg):
```
thibaudcolas, line 25: Let’s rename this to something else like `hreflang_url` to reflect the fact this isn’t meant for canonical links.

thibaudcolas (changes requested):
@Hitansh-Shah here are suggested changes. When opening a pull request like this, please take a moment to describe the changes it contains – what you’ve done, why, and any specific testing steps.

apps/core/templatetags/core_tags.py
```diff
@@ -20,3 +20,8 @@ def navigation(context):
         "current_page": context.get("page"),
         "annotated_list": Page.get_annotated_list_qs(pages),
     }
+
+
+@register.filter
+def hreflang_url(value, arg):
+    return value.replace(arg, "en-latest")
```
laymonage, line 27: Would this cause problems if the page's path has its own `language_code` as a substring? e.g. for a page with the slug `upgrade-latest-version` under the Deutsch latest (`de-latest`) locale, this tag will return something like `/en-latest/upgraen-latest-version`.

It's somewhat of an edge case, but maybe there's something we can do to prevent this (e.g. maybe replacing `/{arg}/` with `/en-latest/` instead of just `arg` with `en-latest`, or alternatively we could also get the `full_url` of the en-latest version of the page).

Hitansh-Shah: Nice catch @laymonage! I think we can just specify `count` parameter in `replace`. For eg `value.replace(arg, "en-latest", 1)`. This will replace only the first instance of the given string. And `/de-latest/` will always appear before any other instance like `upgrade-latest-version` slug like you mentioned above.

laymonage: Yep, I think that works too!

laymonage (commented):
I think this looks good! Would you like to write tests for this?

Hitansh-Shah:
> I think this looks good! Would you like to write tests for this?

Sure, though I don't have much experience in writing tests so it would be nice if you could guide me for this.

laymonage:
> Sure, though I don't have much experience in writing tests so it would be nice if you could guide me for this.

I can think of at least two possible tests:

**Test 1**
- Create a page with a locale other than `en-latest`
- Render the page
- Assert that the `<link rel="alternate" hreflang="en" ...` element exists

**Test 2**
- Create a page with the `en-latest` locale
- Render the page
- Assert that the `<link rel="alternate" hreflang="en" ...` element does not exist

The existing tests in this project such as [this](https://github.com/wagtail/guide/blob/main/apps/core/tests/test_locales.py) and [this](https://github.com/wagtail/guide/blob/main/apps/core/tests/test_page_response.py) should give you the basic idea on how to setup pages and do assertions.

You can use `assertContains` or you can make the test class extend `wagtail.test.utils.WagtailTestUtils` and use `assertTagInHTML`. [Example in Wagtail](https://github.com/wagtail/wagtail/blob/6a11ae711a86e9fa2737cadce72c6a9c622801a4/wagtail/admin/tests/pages/test_explorer_view.py#L734-L750).

Hitansh-Shah:
Hi @laymonage , I have added some tests for `alternate` link tag. Please have a look.

laymonage (approved):

PR #35, opened 2022-08-04
add live search

Hitansh-Shah:
This PR depends on #29 . Also this needs a little more css and js fixing, therefor I have kept it in Draft. Please let me know about any changes and improvements.


vossisboss:
This is a great start @Hitansh-Shah. I'm wondering though if there might be a way for us to capture the parent page name and use that as the "Section" label in the design as well as the meta data description for the article. I'm not 100% sure what the best way to grab that data would be but my inclination would be try to tie it to the meta data description in the Promote tab since that is meant to feed search engines anyway.

Hitansh-Shah:
> This is a great start @Hitansh-Shah. I'm wondering though if there might be a way for us to capture the parent page name and use that as the "Section" label in the design as well as the meta data description for the article. I'm not 100% sure what the best way to grab that data would be but my inclination would be try to tie it to the meta data description in the Promote tab since that is meant to feed search engines anyway.

Yes we can. I am working on it, will push a commit shortly.

Hitansh-Shah:
@vossisboss , I have added parent section and search description as you suggested and as it was in design. I am attaching the screenshot, please have a look.
Screenshot from 2022-08-08 12-52-32

vossisboss:
Hi @Hitansh-Shah. This definitely looks like the design now. Great job! I would like @laymonage or @thibaudcolas to weigh in on the code solution you chose though because I'm curious whether they think adding the djangorestframework package is the best approach or if there is already a native way within Wagtail or Django to return the data needed for the search.

vossisboss (commented):
Hi @Hitansh-Shah. This definitely looks like the design now. Great job! I would like to get @thibaudcolas or @laymonage to get a closer look at your code solution though. I'm not sure whether using the djangorestframework package is the best approach or whether there might be a native way to get the data needed for the search within Wagtail or within Django already.

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,67 @@
...
+    }
+}
+search_input.addEventListener("keyup", onSearchInputChange)
+
+
+const removeExistingChildren = (parent) => {
+    const children = [...parent.children]
+    children.map((child) => {
+        child.remove()
+    })
```
laymonage, line 25: The `map()` method is used to transform a list of objects into another list of objects, so I don't think it's appropriate here. A simple for loop could work?

```suggestion
    for (child of parent.children) {
        child.remove();
    }
```

laymonage (changes requested):
Nice, it seems to work well! I have some minor comments.

Also, a few points:
- It seems that we haven't set up linting for the front-end code (that could be a separate task). It'd be nice to have that in before we start working on more front-end stuff.
- We might be able to get away with just a `JsonResponse` instead of using Django REST Framework, but seeing that you've already got it working with DRF, I don't mind!
- I highly recommend reading [Wagtail's contributing guide](https://docs.wagtail.org/en/stable/contributing/index.html). I don't know if we want to apply them to this project as well, but I think the conventions mentioned there are good for consistency and code readability.

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,67 @@
...
+        const res = await fetch(window.location.origin + '/search_json/?query=' + query)
+        const data = await res.json()
+        console.log(data)
+        injectResultsInHTML(data)
+    } catch(err) {
+        console.log(err)
+        return {"error": err}
+    }
+}
+search_input.addEventListener("keyup", onSearchInputChange)
```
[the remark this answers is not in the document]
Hitansh-Shah: Yeah that sounds reasonable. Should I push a commit for that in this PR or open a separate PR?

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,67 @@
...
+
+const removeExistingChildren = (parent) => {
+    const children = [...parent.children]
+    children.map((child) => {
+        child.remove()
+    })
+}
+searchIconButton.addEventListener("click", () => {
+    const resultsDiv = document.getElementById("results")
+    const resultsCountContainer = document.getElementById("results-count-container")
```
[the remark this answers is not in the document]
Hitansh-Shah: Ohh okay. But I don't understand how we can use data attributes here. For using data attributes, we would first need to fetch dom element using id or class anyway, right? I may have misunderstood your point, please correct me if I am wrong.

Hitansh-Shah:
Thanks @laymonage for the review. I have committed the changes you suggested. I have also commented my doubts in a couple of suggested changes. Please have a look.

apps/frontend/static_src/js/main.js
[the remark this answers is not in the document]
laymonage: Data attributes can also be used for querying the DOM element with `document.querySelector`, e.g.

```js
const resultsDiv = document.querySelector('[data-search-results]');
const resultsCount = document.querySelector('[data-search-results-count]');
```

apps/frontend/static_src/js/main.js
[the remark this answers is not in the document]
laymonage: If you could include it here, that'd be nice, but if it takes time we can leave that for a separate PR.

apps/frontend/static_src/js/main.js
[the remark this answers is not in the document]
Hitansh-Shah: Ohh okay got it, thanks :+1: . I will push a commit for this and implementing delay.

Hitansh-Shah:
@laymonage I have done the requested changes, please have a look now.

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,70 @@
...
+    } catch(err) {
+        console.log(err)
+        return {"error": err}
+    }
+}
+searchInput.addEventListener("keyup", (event) => {
+    setTimeout(() => {
+        onSearchInputChange(event)
+    }, 1000)
+})
```
laymonage, line 21: This seems to only delay the `fetch()` without preventing duplicate requests:

image

Are you familiar with [debouncing/throttling](https://css-tricks.com/debouncing-throttling-explained-examples/)? I think that could help eliminate the duplicate requests.

apps/search/views.py
```diff
@@ -35,3 +38,37 @@ def search(request):
...
+            return ancestors[2].title
+        else:
+            return "Home"
+
+
+@api_view(["GET"])
+def search_json(request):
+
+    search_query = request.GET.get("query", None)
+    # page = request.GET.get("page", 1)
```
laymonage, line 62: Could we remove this if it's unused?

requirements/base.txt
```diff
@@ -1,3 +1,4 @@
 Django>=4.0,<4.1
 wagtail>=3.0,<3.1
 django-manifest-loader
+djangorestframework
```
laymonage, line 4: As we discussed in the meeting, we should add the version specifier here.

apps/core/templates/components/header.html
```diff
@@ -10,15 +10,15 @@
...
           <img src="{% manifest 'images/search.svg' %}" >
         </button>
-        <div class="modal fade" id="searchModal" tabindex="-1" aria-labelledby="searchModalLabel" aria-hidden="true">
+        <div class="modal fade" id="search-modal" tabindex="-1" aria-labelledby="searchModalLabel" aria-hidden="true">
           <div class="modal-dialog modal-lg">
             <div class="modal-content rounded-3 fs-6">
               <form action="{% url 'search' %}" method="get">
                 <div class="modal-body p-4 d-flex align-items-center justify-content-center">
-                  <input type="text" name="query" placeholder="Search..." class="border border-1 border-dark bg-transparent form-control form-control-lg rounded-start w-75 ps-3">
+                  <input type="text" name="query" data-search-input placeholder="Search..." class="border border-1 border-dark bg-transparent form-control form-control-lg rounded-start w-75 ps-3">
```
laymonage, line 21: Could we add a margin between this and the search button?

laymonage (commented):
Hey, nice work with the updates! I had another look at this and found some other rooms for improvements.

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,70 @@
 import { Dropdown } from 'bootstrap';
+
+const searchInput = document.querySelector('[data-search-input]')
+const searchIconButton = document.querySelector('[data-search-icon-button]')
+
+const onSearchInputChange = async (event) => {
+    const query = event.target.value
+    try {
+        const res = await fetch(window.location.origin + '/search_json/?query=' + query)
```
thibaudcolas, line 9: Not sure this matters here but we want to be careful with how we handle input like this, always URL-encoding it as needed.

```suggestion
        const res = await fetch(`/search_json/?${(new URLSearchParams({ query })).toString()}`)
```

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,70 @@
...
+        const data = await res.json()
+        injectResultsInHTML(data)
+    } catch(err) {
+        console.log(err)
+        return {"error": err}
```
thibaudcolas, line 14: Unless I’m misunderstanding the code, this return value doesn’t seem to be used anywhere. Could we instead use `window.alert` to show an error message for end users?

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,70 @@
...
+    }, 1000)
+})
+
+
+const removeExistingChildren = (parent) => {
+    const children = [...parent.children]
+    for (const child of children) {
+        child.remove()
+    }
+}
```
thibaudcolas, line 29:
```suggestion
const removeExistingChildren = (parent) => {
    resultsDiv.innerHTML = ''
}
```

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,70 @@
...
+    resultsCountDiv.appendChild(resultsCountData)
+    resultsCountDiv.classList.add('m-3', 'mx-5')
+    resultsCountContainer.appendChild(resultsCountDiv)
+
+    for(const result of results) {
+        const resultDiv = document.createElement("div")
+        const resultLink = document.createElement("a")
+        const resultDescription = document.createElement("div")
+        const resultParentSection = document.createElement("div")
+        resultLink.appendChild(document.createTextNode(result.title))
```
thibaudcolas, line 56: A bit simpler for a case like this (and below):

```suggestion
        resultLink.innerText = result.title
```

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,70 @@
...
+        resultDescription.appendChild(document.createTextNode(result.search_description))
+        resultParentSection.appendChild(document.createTextNode(result.parent_section))
+        resultLink.href = result.url_path.replace('home-x/', '').replace('home-x-', '')
```
thibaudcolas, line 59: This looks a bit odd to me. I would expect something like this to happen server-side, or if we want to keep it here, then at least add a comment.

thibaudcolas (changes requested):

apps/frontend/static_src/js/main.js
Hitansh-Shah: Nope I am not familiar with that. I will read about that and will try to implement it.

apps/search/views.py
Hitansh-Shah: Sure, I will remove that.

requirements/base.txt
Hitansh-Shah: :+1:

apps/core/templates/components/header.html
Hitansh-Shah: We can. I have just followed the design so I don't know if we should add a margin or not. Maybe @vossisboss and @thibaudcolas can comment on this.

apps/frontend/static_src/js/main.js
Hitansh-Shah: We have language/version slug in every page. Therefore `/search_json/` would return `/en-latest/search_json` instead of `/search_json` which is what we need. For this reason I have used `window.location.origin`.
Apart from this I will add URL-encoding as you suggested above.

apps/frontend/static_src/js/main.js
Hitansh-Shah: Yeah you are right. Initially I didn't know what to do if error occurs, therefore I just logged it out to debug.

apps/frontend/static_src/js/main.js
Hitansh-Shah: This was something confusing to me as well and I wanted to ask about this but it just slipped out of my mind :sweat_smile: .
Here we have a home page with slug `home-x`. However on the site the slug isn't used for home or root page (as far as I know). But in the search results the slug is present.
So `/` would be returned as `/home-x`, `/tutorial/` as `/home-x/tutorial/` and so on.

Hitansh-Shah: As a result when you click on any of the search-results page, you would get 404.

Hitansh-Shah:
@thibaudcolas @laymonage . I have done most of the requested changes, you can re-review it.
1) For debouncing as sage suggested, I added `apps/frontend/static_src/js/lodash_debounce/lodash.custom.min.js` and included it in `main.js`. This allowed to call the `_.debounce()` method.
2) I have specified the version for `djangorestframework`.

I still have some doubts for some changes and I have left a comment for them, please check them out.

apps/core/templates/components/header.html
vossisboss: I'm not sure if Ben took the focus box into account when making the design. So I'm fine with us seeing what it looks like with some margin @Hitansh-Shah.

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,63 @@
 import { Dropdown } from 'bootstrap';
+require('./lodash_debounce/lodash.custom.min.js')
```
laymonage, line 2: I'm a bit surprised we're adding this manually 🤔 

Can't we install it through NPM and `import` it?

Hitansh-Shah: Ohh we can. We only needed the `debounce` function of lodash, so I thought installing the whole `lodash` package would not be necessary and hence I included the custom build which enables us to use debounce.
However it turns out there is also a npm package which allows us to do the same https://www.npmjs.com/package/lodash.debounce
I will change it to use npm package.

Hitansh-Shah:
@laymonage please have a look now.

apps/frontend/static_src/js/main.js
```diff
@@ -1 +1,63 @@
 import { Dropdown } from 'bootstrap';
+const debounce = require('lodash.debounce')
```
laymonage, line 2: Let's use the newer `import` syntax here:

```suggestion
import debounce from 'lodash.debounce';
```

apps/search/views.py
```diff
@@ -35,3 +38,36 @@ def search(request):
...
+        ancestors = page.get_ancestors()
+        if len(ancestors) >= 3:
+            return ancestors[2].title
+        else:
+            return "Home"
+
+
+@api_view(["GET"])
+def search_json(request):
+
```
laymonage, line 60: Remove the empty line?

laymonage (commented):
Some minor comments from me, but otherwise it's looking better! I tested the debouncing and it works, though maybe we want to lower the delay a bit (something like `350`?).

I don't have any major concerns left, so after you're done with the changes I'm fine if we want to merge this to unblock you.

apps/search/views.py
Hitansh-Shah: I always run `make format`. Shouldn't that take care of this?

Hitansh-Shah:
@laymonage I have done the final changes and I have also added some margin as you suggested.

PR #40, opened 2022-08-09
Split CI setup into separate jobs for tests & linting

Hitansh-Shah:
Firstly I have edited Makefile's `test` command to only run tests and not linting. Then I have edited `test.yml` to have separate jobs for testing and linting.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,3 +35,24 @@ jobs:
       - name: Test
         run: |
           make test
+          
+  run-linting:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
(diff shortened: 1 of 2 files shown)


laymonage (approved):
This looks good to me and it seems to work well, nice work!

I think it would be nice to have caching for both [pip](https://github.com/actions/setup-python#caching-packages-dependencies) and [yarn](https://github.com/actions/setup-node#caching-global-packages-data) (or NPM if we decide to go with that), but that should probably be in a separate PR.

Hitansh-Shah:
Thanks @laymonage. Should I merge this then?

vossisboss (approved):
This looks good to me. I say let's merge it and see how well it works.

PR #41, opened 2022-08-10
add previous and next buttons

Hitansh-Shah:
I have added previous and next buttons for content pages. For that I have registered 2 filters namely `previous_page` and `next_page` which have their own logic for returning their respective pages. I have designed the buttons according to the design on figma. I have also added a new scss file `nav_button.scss` in `apps/frontend/static_src/scss/components/` for styling and adding hover effects on the buttons. The buttons seemed to be working fine when tested locally.
I am attaching the screenshot of the end result:
Screenshot from 2022-08-10 23-18-23

### DIFF
--- a/apps/core/templatetags/core_tags.py
+++ b/apps/core/templatetags/core_tags.py
@@ -20,3 +20,42 @@ def navigation(context):
         "current_page": context.get("page"),
         "annotated_list": Page.get_annotated_list_qs(pages),
     }
+
+
+@register.filter
+def next_page(page):
+    first_descendant = page.get_descendants().live().first()
+    if first_descendant:
+        return first_descendant
+
+    next_sibling = page.get_siblings().live().filter(path__gt=page.path).first()
+    if next_sibling:
+        return next_sibling
+
+    next_section = (
+        page.get_ancestors()
+        .live()
+        .last()
+        .get_siblings()
+        .filter(path__gt=page.path)
+        .first()
+    )
+    if next_section:
+        return next_section
+
+    return None
+
+
+@register.filter
+def previous_page(page):
+    previous_page = page.get_siblings().live().filter(path__lt=page.path).last()
+    if previous_page:
+        last_descendant = previous_page.get_descendants().live().last()
+        if last_descendant:
+            return last_descendant
+        return previous_page
+    parent = page.get_ancestors().live().last()
+    if parent:
+        return parent
+
+    return None
--- a/apps/core/templates/core/content_page.html
+++ b/apps/core/templates/core/content_page.html
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 
-{% load wagtailcore_tags core_tags %}
+{% load wagtailcore_tags core_tags manifest %}
 
 {% block content %}
 <div class="d-none d-md-block col-md-4 col-lg-3 pe-md-5">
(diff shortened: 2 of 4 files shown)


vossisboss (changes requested):
Hi @Hitansh-Shah. I'm of the opinion that on the general category pages, it makes sense to give folks a button to go to the first descendant within the category, but returning to the last page or the last descendant won't be very helpful for users. I think this navigation will be most helpful for helping them click between pages within a category. If they want to move to a different category entirely, then they can use the lefthand menu to navigate to their preferred section. Let me know what you think.

@laymonage I feel like Hitansh's filter might be the best approach for this type of site in which we aren't creating a model for every single page category. If you know of other methods we should consider, please let us know.

apps/core/templatetags/core_tags.py
```diff
@@ -20,3 +20,41 @@ def navigation(context):
...
+        return (
+            page.get_ancestors()
+            .live()
+            .last()
+            .get_siblings()
+            .filter(path__gt=page.path)
+            .first()
+        )
+    else:
+        return None
```
laymonage, line 42: Note that each `get_descendants()` and `get_siblings()` call returns a QuerySet, and using it directly in an `if` expression would evaluate the queryset (i.e. execute the query), returning all objects that match the query. This means that there would be some duplicated queries.

Possible alternative:

```suggestion
@register.filter
def next_page(page):
    first_descendant = page.get_descendants().live().first()
    if first_descendant:
        return first_descendant

    next_sibling = page.get_siblings().live().filter(path__gt=page.path).first()
    if next_sibling:
        return next_sibling

    next_section = (
        page.get_ancestors()
        .live()
        .last()
        .get_siblings()
        .filter(path__gt=page.path)
        .first()
    )
    if next_section:
        return next_section

    return None
```

apps/core/templatetags/core_tags.py
```diff
@@ -20,3 +20,41 @@ def navigation(context):
...
+            .live()
+            .filter(path__lt=page.path)
+            .order_by("-path")
+            .first()
+        )
+        if prev_page.get_descendants().live():
+            return prev_page.get_descendants().live().last()
+        return prev_page
+    else:
+        return page.get_ancestors().live().last()
```
laymonage, line 60: Similarly,

```suggestion
@register.filter
def previous_page(page):
    previous_page = page.get_siblings().live().filter(path__lt=page.path).last()
    if previous_page:
        last_descendant = previous_page.get_descendants().live().last()
        if last_descendant:
            return last_descendant
        return previous_page

    parent = page.get_ancestors().live().last()
    if parent:
        return parent

    return None
```

laymonage (changes requested):

apps/core/templatetags/core_tags.py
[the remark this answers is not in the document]
laymonage: > I'm of the opinion that on the general category pages, it makes sense to give folks a button to go to the first descendant within the category, but returning to the last page or the last descendant won't be very helpful for users.

Honestly, I think it's fine. Wagtail's docs also does this, and I've found other docs that have the same behaviour (e.g. [Tailwind](https://tailwindcss.com/docs/), [new React docs](https://beta.reactjs.org/learn), [Vue](https://vuejs.org/guide/)).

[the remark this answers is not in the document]
Hitansh-Shah: Yes, I also implemented the logic keeping Wagtail's docs in mind. However I am fine with both the styles. Though in the above suggested change, in `previous_page` there is 
```python
if page.get_descendants():
        return None
```
this can cause confusion when there are multiple levels of nesting inside any particular category like `Tutorial, etc.`

apps/core/templatetags/core_tags.py
Hitansh-Shah: Oh I see... Thanks for the suggestion, I will update the PR.

laymonage (approved):
I'm happy with this! Again, would be nice to have tests, but we can do that separately if you'd like to learn more on testing before writing them.

Hitansh-Shah:
Cool, I guess we can merge it. I will open a separate PR for tests as it is listed as a separate task.

PR #42, opened 2022-08-12
add kumbh-sans font

Hitansh-Shah:
I have added custom font (Kumbh-Sans). I have included the font file locally so we do not have to load the font from google.


apps/frontend/static_src/scss/variables.scss
```diff
@@ -9,6 +9,7 @@ $body-bg: #f5f3f8;
 $font-size-root:             null !default;
 $font-size-base:             1rem !default; // Assumes the browser default, typically `16px`
 $font-size-sm:               $font-size-base * .875 !default;
+$font-family-base:           Kumbh Sans;
```
vossisboss, line 12:
```suggestion
$font-family-base:           Kumbh Sans, Arial, Helvetica, sans-serif;
```

vossisboss (changes requested):
Hi @Hitansh-Shah. Two small things to address on this one. First, even when loading fonts locally, you should always include a couple back-up fonts that the browser can fall back to if the custom font file doesn't load for some reason. Second, could you please create a fonts file in the `static_src` directory and put the font file there instead of the `scss` directory. I think that would help people find the font files a bit more easily in the future. With those changes, this PR should be pretty much ready to merge.

apps/frontend/static_src/scss/variables.scss
thibaudcolas: Here are Wagtail’s fallback fonts:

```suggestion
$font-family-base:           Kumbh Sans, -apple-system, BlinkMacSystemFont, "Segoe UI", system-ui, Roboto, "Helvetica Neue", Arial, sans-serif, Apple Color Emoji, "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
```

Link: https://github.com/wagtail/wagtail/blob/main/client/src/tokens/typography.js#L11

apps/frontend/static_src/scss/kumbh_sans.scss
```diff
@@ -0,0 +1,16 @@
...
+  unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
+}
+/* latin */
+@font-face {
+  font-family: 'Kumbh Sans';
+  font-style: normal;
+  font-weight: 400;
+  src: url(./font/KumbhSans-VariableFont_wght.woff) format('woff2');
+  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+}
```
thibaudcolas, line 16:
```suggestion
/* latin and latin-ext */
@font-face {
  font-family: 'Kumbh Sans';
  font-style: normal;
  font-weight: 400;
  src: url(./font/KumbhSans-VariableFont_wght.woff2) format('woff2');
}
```

Since this is the same variable font file used for both the latin and latin-ext character sets, there’s nothing to gain by defining this separately with [`unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) (it’s only useful to decide which font file to download).

thibaudcolas (changes requested):
@Hitansh-Shah I’ve added two comments. In addition, make sure to rename the font file to the `.woff2` extension so it’s clearer we’re using the correct format.

Hitansh-Shah:
Thanks @thibaudcolas  and @vossisboss for the review. I will update the PR soon.

Hitansh-Shah:
@vossisboss @thibaudcolas , I have done the requested changes, please have a look.

PR #44, opened 2022-08-13
add tests for previous_page and next_page template filters

Hitansh-Shah:


apps/core/tests/test_previous_page_and_next_page.py
```diff
@@ -0,0 +1,32 @@
...
+        self.ab = Page(title="ab", slug="ab", show_in_menus=True)
+        self.a.add_child(instance=self.ab)
+        self.ac = Page(title="ac", slug="ac", show_in_menus=True)
+        self.a.add_child(instance=self.ac)
+        self.b = Page(title="b", slug="b", show_in_menus=True)
+        self.home.add_child(instance=self.b)
+
+    def test_previous_page_filter(self):
+        self.assertEqual(previous_page(self.a).specific, self.home)
+        self.assertEqual(previous_page(self.ab), self.a)
```
thibaudcolas, line 23: Isn’t this the same test as the one before?

thibaudcolas (changes requested):
The assertions themselves are looking spot on @Hitansh-Shah, but I think this needs more work on the test structure. The issues I see are that:

- It’s hard to understand exactly what cases are tested
- With the short and similar-looking variable names, it’s hard to keep in mind which variable is what
- Since the pages to test are defined in the test setUp method, you have to go back and forth between it and the two other methods.

I’d recommend using more descriptive variable names, and adding comments:

```python
        self.first_child = Page(title="a", slug="a", show_in_menus=True)
        self.home.add_child(instance=self.first_child)
        self.subchild1 = Page(title="ab", slug="ab", show_in_menus=True)
        self.first_child.add_child(instance=self.subchild1)
        self.subchild2 = Page(title="ac", slug="ac", show_in_menus=True)
        self.first_child.add_child(instance=self.subchild2)
        self.last_child = Page(title="b", slug="b", show_in_menus=True)
        self.home.add_child(instance=self.last_child)

    def test_previous_page_filter(self):
        # A page’s first child links to its parent as previous page.
        self.assertEqual(previous_page(self.first_child).specific, self.home)
        # A page’s first child links to its parent as previous page.
        self.assertEqual(previous_page(self.subchild1), self.first_child)
        # If there is a previous sibling, we link to it.
        self.assertEqual(previous_page(self.subchild2), self.subchild1)
        # If a sibling page has child pages, we link to the sibling’s last child.
        self.assertEqual(previous_page(self.last_child), self.subchild2)
```

Alternatively you could also do this by adding separate test methods for each of the cases:

```python
    def test_previous_without_sibling_goes_to_parent(self):
        self.assertEqual(previous_page(self.first_child).specific, self.home)
```

This is more verbose, but is preferred sometimes since it makes it very clear where one test case stops and the other starts. I think it’s a bit overkill in this instance but worth considering.

apps/core/tests/test_previous_page_and_next_page.py
Hitansh-Shah: Are you talking about `test_previous_page_filter` and `test_next_page_filter`? I don't understand which test are you talking about.

apps/core/tests/test_previous_page_and_next_page.py
```diff
@@ -0,0 +1,32 @@
+from django.test import TestCase
+from wagtail.core.models import Page
+
+from apps.core.factories import HomePageFactory
+
+from ..templatetags.core_tags import next_page, previous_page
```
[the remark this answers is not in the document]
Hitansh-Shah: Cool :+1:

Hitansh-Shah:
Thanks @thibaudcolas for the review, I will update the PR soon. Also I didn't understand the comment about "same test". Please elaborate on that one.

apps/core/tests/test_previous_page_and_next_page.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class TestNavigation(TestCase):
+    def setUp(self):
+        self.home = HomePageFactory()
+        self.a = Page(title="a", slug="a", show_in_menus=True)
```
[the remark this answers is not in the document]
Hitansh-Shah: Just one doubt. The pages created in the setUp function should have `show_in_menus=True`. Is there a way to do this when using `ContentPageFactory`?

apps/core/tests/test_previous_page_and_next_page.py
Hitansh-Shah: Ohh okay I got it. I added those tests so that it covers every type of page (home and content). Yeah the logic is same for both of them and we can remove one if we want.

Hitansh-Shah:
@thibaudcolas I have done the suggested changes, please have a look. Also I had doubts in some of them for which I have left comments, please check them out.

PR #45, opened 2022-08-19
Test coverage

Hitansh-Shah:
This PR depends on #44 .
I installed coverage package as a dev dependency and ran tests with it. According to its report we have a total of 94% test coverage. I am attaching the full report.
image


thibaudcolas (changes requested):
@Hitansh-Shah upon further review the results you’re sharing seem correct to me. 

To fully complete this, you will need to:

- Add an ignore pattern to `.gitignore` so we don’t have the `.coverage` file committed
- Either update the existing `make test` command in the Makefile, or make a new `make test-coverage` command, running `coverage run manage.py test && coverage report`
- Then (if needed) update our GitHub Actions workflow to run tests with this updated command
- Add a bit more configuration for coverage collection to our setup.cfg

In particular on configuration, I would recommend adding:

```
[coverage:run]
source = .
omit =
   manage.py
   *venv/*
   *node_modules*
   *migrations*

[coverage:report]
show_missing = True
```

This will clean up the report a bit.

Hitansh-Shah:
Thanks @thibaudcolas . I have done the requested changes.

setup.cfg
```diff
@@ -10,3 +10,14 @@ skip_glob=
     */migrations/*
 profile = black
 known_first_party = apps
+
+[coverage:run]
+source = .
+omit =
+   manage.py
+   *venv/*
```
allcaps, line 18: Why does this have a `/` and other dirs don't? I guess we can remove `/`.

Can we add `*env*` as our own instructions tell to create a env by this name?

Can we add `*test*` as tests should not be in coverage.

Hitansh-Shah: Yup, seems fair to me. I have pushed a commit for this.

PR #47, opened 2022-08-22
Heroku environment

RealOrangeOne:
This PR adds the required configuration to setup a staging environment (and in future, production environment).

Much of this configuration is duplicated from [wagtail-kit](https://git.torchbox.com/internal/wagtail-kit) (torchbox-only link)


vossisboss (commented):
Hello @RealOrangeOne. I had a look through all these files. Only thing I spotted that obviously needed a fix was a typo in a comment that I just couldn't resist correcting.

I have a couple other general questions though:

1. You added some configuration for Postgres but didn't update the database settings to include Postgres yet. Is this something that typically happens later closer to pushing to production? Our did you just not include it because Hitansh is still using the default sqlite database?
2. I'm newer to caching options, so I'm curious why you have database caching set up as a backup to Redis. Is database caching the bare minimum type of caching that you would recommend for a Wagtail site?
3. Some of these configuarations are clearly going to change how @Hitansh-Shah and the rest of us develop this project locally. How do you typically recommend handling that? With Docker? Or a local.py settings file? 
4. If we use local settings, how is Redis typically handled? Is Redis for staging and production only typically? Or is there some benefit to having it set up in a local environment?

RealOrangeOne:
Great questions!

> You added some configuration for Postgres but didn't update the database settings to include Postgres yet

I have. Where [`DATABASES`](https://github.com/wagtail/guide/pull/47/files#diff-027ab3a67434a9d91da9f1c81abbf12ab645734dc5473851d18340eab66fa057R105) is defined, it now uses `dj-database-url`, which magically creates the required settings based on a single `DATABASE_URL` environment variable. This way we can use postgres in GitHub Actions and Heroku, but without impacting local development (thanks to the `default` argument).

>  I'm curious why you have database caching set up as a backup to Redis

Caching is massively helpful for a lot of performance things, especially around the admin. I can definitely see an argument for having no cache when Redis isn't available, but left it like this to more closely match Torchbox's standard project structure. This just means that when redis isn't configured, there's still something cache-like to use.

> Some of these configuarations are clearly going to change how Hitansh-Shah and the rest of us develop this project

I've intentionally tried to keep these to a minimum. If you could check out the branch locally, try running it and letting me know if / what has changed massively, I'm sure I can tweak things slightly to make them a lot easier.

> Is Redis for staging and production only typically

Exactly! We usually have some form of cache enabled locally for performance. Functionally there's no difference to Redis vs another cache, just the performance. Redis is typically faster than a database for its cache, but locally it means we only need a database and don't need a separate redis instance. That's helped even more by the fact we can keep using sqlite as the database locally, and store our cache with it too, with 0 extra components running.

apps/guide/settings/base.py
```diff
@@ -184,4 +238,228 @@
...
+#
+# Three required environment variables are:
+#  * AWS_STORAGE_BUCKET_NAME
+#  * AWS_ACCESS_KEY_ID
+#  * AWS_SECRET_ACCESS_KEY
+# The last two are picked up by boto3:
+# https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variables
+if "AWS_STORAGE_BUCKET_NAME" in env:
+    # Add django-storages to the installed apps
+    INSTALLED_APPS = INSTALLED_APPS + ["storages", "wagtail_storages"]
```
zerolab, line 262: wagtail_storages is only needed if we're to have private documents. Given this is a public project, is it really needed?

apps/guide/settings/base.py
```diff
@@ -184,4 +238,228 @@
...
+
+    # Do not allow overriding files on S3 as per Wagtail docs recommendation:
+    # https://docs.wagtail.io/en/stable/advanced_topics/deploying.html#cloud-storage
+    # Not having this setting may have consequences in losing files.
+    AWS_S3_FILE_OVERWRITE = False
+
+    # Default ACL for new files should be "private" - not accessible to the
+    # public. Images should be made available to public via the bucket policy,
+    # where the documents should use wagtail-storages.
+    AWS_DEFAULT_ACL = "private"
```
zerolab, line 281: note: without wagtail_storages, this doesn't need to be private

apps/guide/settings/base.py
```diff
@@ -184,4 +238,228 @@
...
+    if "SENTRY_ENVIRONMENT" in env:
+        sentry_kwargs.update({"environment": env["SENTRY_ENVIRONMENT"]})
+
+    release = get_default_release()
+    if release is None:
+        try:
+            # But if it's not, we assume that the commit hash is available in
+            # the GIT_REV environment variable. It's a default environment
+            # variable used on Dokku:
+            # http://dokku.viewdocs.io/dokku/deployment/methods/git/#configuring-the-git_rev-environment-variable
```
zerolab, line 416: nitpick: dokku

apps/guide/settings/dev.py
```diff
@@ -1,7 +1,7 @@
 from .base import *  # noqa
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = False
```
zerolab, line 4: question: why not? this is the dev settings file

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 88
-max-complexity = 6
+max-complexity = 8
```
zerolab, line 3: I am curious about this change. did the tests not pick it up before?

Dockerfile
```diff
@@ -1,60 +1,83 @@
...
 
-# Set environment variables.
-# 1. Force Python stdout and stderr streams to be unbuffered.
-# 2. Set PORT variable that is used by Gunicorn. This should match "EXPOSE"
-#    command.
-ENV PYTHONUNBUFFERED=1 \
-    PORT=8000
-
-# Install system packages required by Wagtail and Django.
-RUN apt-get update --yes --quiet && apt-get install --yes --quiet --no-install-recommends \
```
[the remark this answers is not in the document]
RealOrangeOne: It might, but I'd want to be completely sure. We've been running the full-fat debian images in production for years without issue. Whilst it'd be nice to know whether this works 100% with what we need, I'd also want to test beforehand.

Although, perhaps this is the ideal testing ground :thinking:

apps/guide/settings/base.py
RealOrangeOne: It still probably should be, so we rely on the bucket policy allocating permissions, not something in the files themselves.

apps/guide/settings/dev.py
RealOrangeOne: I have literally 0 recollection of changing that :sweat_smile:

setup.cfg
RealOrangeOne: The problem was `if "SENTRY_DSN" in env and not is_in_shell:` was considered too complex - which just seems silly

apps/guide/settings/base.py
zerolab: works for me

setup.cfg
zerolab: perhaps that line should get a `# noqa: C901` ? 🤔

RealOrangeOne:
> should we switch local development to Postgres or just keep going with sqlite

Let's keep using sqlite. I'd be very very shocked if there were any issues as a result of that. Now we're testing with postgres in GitHub Actions, that should catch any potential issues anyway.

> So what you're saying is that if we don't have Redis set up locally, then the settings will default to caching in the local database without breaking our local environment?

Precisely!

setup.cfg
RealOrangeOne: Perhaps, but then also i'd argue that line is definitely at a level of complexity which is fine in a codebase, hence changing the rule rather than special-casing that specific line.

vossisboss:
The branch is working for me now. I'm inclined to give Hitansh a chance to test it too before we merge. If he doesn't have time today, I'll make sure we cover it at our checkin tomorrow.

vossisboss:
Does `python manage.py collectstatic` resolve the error for you @Hitansh-Shah?

If so, we might need to add that to the make steps.

Hitansh-Shah:
Yes it worked!

RealOrangeOne:
Fixed in https://github.com/wagtail/guide/pull/47/commits/2115d0e9baddf49910435f0e6a69638a16120852, so that shouldn't be necessary anymore. You can test the fix worked by removing the `static/` directory in the root of the project and re-running tests.

Hitansh-Shah (approved):
I tested after the latest commit and I was able to successfully run all the `makefile` commands. I am still very much learning about deployment and environment setup for relatively big and complex sites like this one itself so I can't really comment on the code :sweat_smile: as of now.

vossisboss (approved):
@zerolab Any other thoughts on this one? If not, we're going to merge this.

apps/guide/settings/base.py
zerolab: I am still nitpicking on this as I don't see the need for it.
less stuff to install/load, the better :P

thibaudcolas: Added as an issue to follow up on later: https://github.com/wagtail/guide/issues/54.

RealOrangeOne:
Automatic deploy has now been configured for `main`

PR #48, opened 2022-08-24
add robots.txt and sitemap.xml

Hitansh-Shah:
I have followed [this](https://www.accordbox.com/blog/wagtail-seo-guide/) tutorial to add `robots.txt` and `sitemap.xml` in the project.

### DIFF
--- a/apps/core/templates/robots.txt
+++ b/apps/core/templates/robots.txt
@@ -1,0 +1,6 @@
+User-Agent: *
+Disallow: /admin/
+Disallow: /search/
+Disallow: /search_json/
+
+Sitemap: {{ wagtail_site.root_url }}/sitemap.xml
--- a/apps/guide/views.py
+++ b/apps/guide/views.py
@@ -1,0 +1,14 @@
+from django.views.generic import TemplateView
+from wagtail.core.models import Site
+
+
+class RobotsView(TemplateView):
+
+    content_type = "text/plain"
+    template_name = "robots.txt"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        request = context["view"].request
+        context["wagtail_site"] = Site.find_for_request(request)
+        return context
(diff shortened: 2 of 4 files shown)


vossisboss (approved):
This looks good to me. Unless you have any different recommendations for handling these items @allcaps, I think these changes can be merged.

apps/core/templates/robots.txt
```diff
@@ -0,0 +1,6 @@
+User-Agent: *
+Disallow: /admin/
+Disallow: /search/
```
thibaudcolas, line 3: I’m not clear on the value of disallowing this. If we retain a standalone search page, we’d probably want it crawled. If we don’t retain the page, we don’t need it listed here.

apps/core/templates/robots.txt
```diff
@@ -0,0 +1,6 @@
...
+Disallow: /search_json/
+
+Sitemap: {{ wagtail_site.root_url }}/sitemap.xml
```
thibaudcolas, line 6:
```suggestion
Sitemap: {{ wagtail_site.root_url }}/sitemap.xml

```

Missing a newline at the end of this file. Feels a bit overkill to me to have the robots.txt be a TemplateView so we list a sitemap using the default `sitemap.xml`, but I guess it might be helpful if we ended up with lots of pages needing multiple sitemaps.

apps/guide/views.py
```diff
@@ -0,0 +1,14 @@
+from django.views.generic import TemplateView
+from wagtail.core.models import Site
+
+
+class RobotsView(TemplateView):
```
thibaudcolas, line 5: I’m surprised we’d want a view inside the `guide` app, but this feels simple enough to refactor later on so let’s leave this as-is for now.

thibaudcolas (approved):
Looking good! There are some things I feel like could be simpler, but they don’t need addressing now.

PR #51, opened 2022-08-26
add inline code style

Hitansh-Shah:
I have followed [this](https://docs.wagtail.org/en/stable/extending/extending_draftail.html) tutorial to add `inlineCode`. The style is reflected in editor but it is not reflected on the page itself. I don't know what is the problem or what I am doing wrong.
I am attaching some screenshots.
image
image


Hitansh-Shah:
So it was missing some parts. The style defined in `control` is for rich text editor itself. It controls how it will look in the editor. For it to reflect on the page, we need to add some `props` (like class) which defines the style that it would use on the page. This procedure was somewhat stated in the docs but for `blocks` and not in `inline styles` which confused me. 
I have made the changes and it works now :smile: 
image

thibaudcolas (changes requested):

Hitansh-Shah:
@thibaudcolas I have done the requested changes. I am attaching a screenshot of the final result.
image

PR #56, opened 2022-09-01
Prev next

allcaps:
A simplification of prev next.

Pages are ordered by default. You don't need to take parent/child into account.

This is a draft because:
- Current implementation needs to be removed
- Tests adjusted
- Styling applied

Not sure if I'd continue with this, as current code 'works'.


Hitansh-Shah:
Nice catch @allcaps . I think we should update our code to use this logic. Also while looking at the existing code I found that it only works correctly when there are atmost 2 levels of nesting. This is something I didn't notice initially while writing the logic. I will push a commit for this soon.

PR #59, opened 2022-09-03
Add feedback system

Hitansh-Shah:
- I created a model `Feedback` which consists of fields `feedback` (happy or unhappy), `feedback_text` (user can provide some additional feedback through this when unhappy) and `page` (a fk to `Page`).
- I created a view `create_feedback` which creates feedback.
- I added added some frontend. The emoji buttons when clicked send a `POST` request to the above view and a feedback is created. If the user clicks the unhappy emoji button then a modal pops up asking for any additional (**optional**) comments before sending the request. Then, when the user clicks "Finish", a feedback is created by sending a post request to the view.
- I registered `Feedback` model with `modeladmin`. As a result all the feedbacks are accessible and visible through admin. The moderator can filter by `Page` and category of feedback (happy or unhappy). I have also added `feedback_text` and `Page` to the `search_fields`. However it is throwing some kind of error when searched. I am attaching the screenshot below along with the demo.

https://user-images.githubusercontent.com/76250591/188265270-5863fb50-4bb4-483e-91a2-7a3b800be2cb.mp4

image


allcaps:
Thank you. Most of it seems to work as discussed. I like the model admin. We are almost there.

1. If a user hits `:(`, the unhappy vote isn't recorded until the user clicks 'finish'. I don't think that is right. The popup title displays "Thanks for your feedback!". So, we should have recorded the unhappy vote to the db. If the user adds additional text, we should update the existing vote.
2. 'Finish' is a slightly weird button text. What about 'Submit'.
2. I don't like the popup as it blocks the page contents. If you give feedback on something, it is better to have access to the  information you give feedback on. The user might want to read back the text that was not clear and needs improving and maybe even copy/paste a snippet text.
3. We need to keep the CSRF token. The `@csrf_exempt` is bad.
4. I'm not so happy with a stand-alone view. A ContentPage has a view, we should override it to accept post requests. This way we also don't need to lookup the page (the page is the current instance `self`). I will provide an example.

Hitansh-Shah:
Thanks @allcaps for the review,
I will update most of the things you stated. Also I kept `@csrf_exempt` because without it a POST request to the view was giving 403 Forbidden error. I didn't know how to resolve it so I temporarily kept it.
For overriding the `ContentPage` view, should I define our own `serve` method inside the page model?

allcaps:
Google it, you should be able to add the token in js to the post request.

And yes, the serve method.

Hitansh-Shah:
Thanks @allcaps . I added the code for csrf token. I will try adding the serve method now.

Hitansh-Shah:
@allcaps I have done the requested changes. You can re-review it.

Hitansh-Shah:
https://user-images.githubusercontent.com/76250591/188323491-41ad7817-9bf4-455c-9ac8-491cdf5b0865.mp4

apps/core/wagtail_hooks.py
```diff
@@ -0,0 +1,19 @@
...
+    model = Feedback
+    base_url_path = "feedbackadmin"
+    menu_label = "Feedback"
+    menu_icon = "pilcrow"
+    add_to_settings_menu = False
+    exclude_from_explorer = False
+    add_to_admin_menu = True
+    list_display = ("feedback", "feedback_text", "page")
+    list_filter = ("feedback", "page")
+    search_fields = ("feedback_text", "page")
```
allcaps, line 16: search_fields = ("feedback_text", "page__title")

apps/core/wagtail_hooks.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class FeedbackAdmin(ModelAdmin):
+    model = Feedback
+    base_url_path = "feedbackadmin"
+    menu_label = "Feedback"
+    menu_icon = "pilcrow"
+    add_to_settings_menu = False
+    exclude_from_explorer = False
+    add_to_admin_menu = True
```
allcaps, line 13: We can drop these:
```
add_to_settings_menu = False
    exclude_from_explorer = False
    add_to_admin_menu = True
    ```
    as those are the default values.
    
    We can also drop `menu_label = "Feedback"` and add to the model:
    
    ```
    class Meta:
         verbose_name="Feedback"
         verbose_name_plural="Feedback items"
    ```
    
This wil give nicer labels in the admin.

allcaps (changes requested):
Nice one! This is much better.

Just some nitpicks and an easy search fix.

Would you mind to add a test for the view?

Hitansh-Shah:
Thanks @allcaps . I will push the requested changes shortly.
Sure I will try adding tests.

apps/core/wagtail_hooks.py
Hitansh-Shah: Yes! It works now. Thanks

Hitansh-Shah:
Hey @allcaps, I have done the changes you suggested. I have also pushed a commit for test. Please have a look.

allcaps (approved):
Thank you!

allcaps:
@Hitansh-Shah Conflicts :/ Can you fix them? 🙏

Hitansh-Shah:
> @Hitansh-Shah Conflicts :/ Can you fix them? 🙏

Sure.

Hitansh-Shah:
@allcaps Done.

allcaps (approved):

PR #61, opened 2022-09-05
refactor previous and next buttons logic

Hitansh-Shah:
Closes #56 
@allcaps  I have refactored the logic as you suggested. I have also adjusted the frontend and the tests. Please have a look.


apps/core/templates/components/nav_buttons.html
```diff
@@ -1,24 +1,24 @@
 {% load wagtailcore_tags core_tags manifest %}
 
 <div class="d-flex mt-5 flex-wrap justify-content-between">
-    {% if page|previous_page != None %}
-      <a href="{% pageurl page|previous_page %}" class="text-decoration-none">
+    {% if previous != None %}
```
allcaps, line 4: The `!= None` is irrelevant. Just do `{% if previous %}`.

Same for next.

allcaps (changes requested):
Ah, nice!

Can you remove the `!=None` from the template?

Hitansh-Shah:
@allcaps I have pushed a commit for this. I think we can merge this now.

PR #117, opened 2022-10-16
Add inline keyboard text style in rich text. Fix #81

### LINKED ISSUE #81: Add inline text styles for keyboard inputs
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd

Hitansh-Shah:
Fixes #81 
@thibaudcolas I have added the inline style for keyboard inputs. I am attaching the screenshots of the final result. I have copied the css from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd. Let me know if the css is to be changed.
image
image


thibaudcolas (approved):
Looking spot on!

PR #132, opened 2022-10-18
Feature/font rework

nicklee:
Adds a font mixin to apply font sizes based on breakpoint and also automatically convert px to rem.


apps/frontend/static_src/scss/components/heading.scss
```diff
@@ -1,8 +1,17 @@
+// Reset margins
+h1,
+h2,
+h3,
+h4,
+h5 {
```
thibaudcolas, line 6: 😬 

```suggestion
h5,
h6 {
```

apps/frontend/static_src/scss/main.scss
```diff
@@ -1,29 +1,32 @@
...
 @import "~bootstrap/scss/utilities";
 
-// Layout & components
-@import "~bootstrap/scss/root";
-@import "~bootstrap/scss/reboot";
-@import "~bootstrap/scss/type";
+// // Layout & components
+// @import "~bootstrap/scss/root";
+// @import "~bootstrap/scss/reboot";
+// @import "~bootstrap/scss/type";
```
thibaudcolas, line 21: This worries me a bit. I’d have assumed lots / all of the Bootstrap components we are using assume they’re built upon Bootstrap’s `root` and `reboot`. Is there any way we could keep those from Bootstrap instead of replacing with normalize?

If we do want to keep normalize – then I think it’d be nice to use the latest version from https://necolas.github.io/normalize.css/ rather than an older one.

thibaudcolas (changes requested):

apps/frontend/static_src/scss/main.scss
nicklee: yea I agree, I'm hoping to eventually remove all of the bootstrap elements, or at least all of the styling (just leaving the functionality for modals/dropdowns etc that are already implemented). I'll update the normalize for the time being.

PR #130, opened 2022-10-18
Use Poetry and remove wagtail-storages. Fix #77, #54

laymonage:
Fix #77 and fix #54.

### Locally

- [Install Poetry](https://python-poetry.org)
- Clone the project
- Run `make backend && make run`

### On Gitpod

- Launch a new instance
- Verify that the server runs once the workspace is loaded, and virtual environment can be activated using `poetry shell`

### Docker

- Run `docker build .`
- Run the image

PR #143, opened 2022-10-18
Use wagtail-localize and update monkey patches for versioning

### LINKED ISSUE #62: Set up wagtail-localize for translations
Once we definitely want to have translations, we need to set up <https://github.com/wagtail/wagtail-localize>. cc @kaedroho whom we will want to get involved to review our setup.

laymonage:
Fixes #62. Depends on #142.

~~For now, we'll only maintain one 'latest' version.~~

~~Handling multiple versions will be reconsidered in the future.~~

Monkey patches have been updated to better handle languages with regional codes. Version has been updated to include `latest` and `4.1.x`.


apps/guide/settings/base.py
```diff
@@ -264,7 +312,7 @@
 # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variables
 if "AWS_STORAGE_BUCKET_NAME" in env:
     # Add django-storages to the installed apps
-    INSTALLED_APPS = INSTALLED_APPS + ["storages", "wagtail_storages"]
+    INSTALLED_APPS = INSTALLED_APPS + ["storages"]
```
laymonage, line 315: Missed this in #130, sorry

apps/guide/settings/base.py
```diff
@@ -190,12 +190,60 @@
...
+    ("ta", "Tamil"),
+    ("tet", "Tetum (Tetun)"),
+    ("th", "Thai"),
+    ("tr", "Turkish"),
+    ("tr-TR", "Turkish (Turkey)"),
+    ("uk", "Ukrainian"),
+    ("vi", "Vietnamese"),
+    ("zh", "Chinese"),
+    ("zh-Hant", "Chinese Traditional"),
+    ("zh-CN", "Chinese (China)"),
```
allcaps, line 246: Same as my other remark. We need the `-latest` and `-X.X.X` if we want to present versioned content.

apps/core/templates/components/language_selector.html
```diff
@@ -13,7 +13,7 @@
                 {% get_language_info for translation.locale.language_code as lang %}
```
saevarom, line 13: Should we somehow limit the list of languages in the language selector to "actually" translated languages?
The long list of languages does not look too pretty right now...

Screenshot 2022-10-20 at 17 14 30

saevarom: This actually works once I start translating the subpages:

Screenshot 2022-10-20 at 17 35 45

So perhaps this is just some kind of artifact of the local installation having a homepage version generated in each language when I build the fixtures... ?

saevarom (approved):
This looks good to me. I had some issues with the language selector which are probably not relevant since I was looking at some generated homepages on my local installation.

apps/core/templates/components/language_selector.html
laymonage: Ah, yes I think we should limit the fixtures to only generate a subset of the available languages. Two or three, combined with `latest` and `4.1.x` should be enough.

laymonage: I've changed the `buildfixtures` management command to only build for `en`, `nl`, and `is-is` with versions `latest` and `4.1.x`.

allcaps: Why would we use `is-is` instead of `is`? AFAIK the Icelandic spoken on Iceland is 'the default'.

saevarom: Agree with @allcaps , we only use `is`

laymonage: That's what I thought, but Wagtail itself seems to use `is-is`: https://github.com/wagtail/wagtail/tree/main/wagtail/locales/locale

In that case, I guess I should update the others as well, e.g. `id-id`.

laymonage: Done, I've matched the codes to Django's `LANG_INFO`: https://github.com/django/django/blob/main/django/conf/locale/__init__.py

(except `pt` which I kept as `pt-pt` as it's clearer that way, and Django's `get_language_info` will handle this anyway)

saevarom:
I believe you could now merge main into your branch to get the linters to work properly.

apps/core/templatetags/core_tags.py
```diff
@@ -24,4 +24,9 @@ def navigation(context):
 
 @register.filter
 def hreflang_url(value, arg):
-    return value.replace(arg, "en-latest", 1)
+    return value.replace(f"/{arg}/", "/en-latest/", 1)
```
thibaudcolas, line 27: Nice one

apps/core/templatetags/core_tags.py
```diff
@@ -24,4 +24,9 @@ def navigation(context):
...
+
+
+@register.simple_tag
+def get_version_from_language_code(language_code):
+    return language_code.rsplit("-", maxsplit=1)[1]
```
thibaudcolas, line 32: TIL about `maxsplit`

thibaudcolas (approved):
Nice one @laymonage. Just gave this a cursory look and test in the admin since this isn’t my area of expertise. All seems well.

PR #148, opened 2022-10-19
Adding focus override utility. Fix #36

### LINKED ISSUE #36: Update focus style to follow WCAG guidelines
The current focus style is #b8cfd5 over #EDEAF3, which isn’t enough contrast ([contrast grid](https://contrast-grid.eightshapes.com/?version=1.1.0&background-colors=&foreground-colors=%23EDEAF3%0D%0A%23B8CFD5&es-color-form__tile-size=compact&es-color-form__show-contrast=aaa&es-color-form__show-contrast=aa&es-color-form__show-contrast=aa18&es-color-form__show-contrast=dnp)) to pass accessibility checks. We need to switch to another color that has at least 3:1 contrast against all backgrounds where the focus would appear.

nicklee:
Adds a global outline to all interactive elements


thibaudcolas (changes requested):
@nicklee code is looking good to me but you forgot to import the file, so I can’t actually test this 😛

nicklee:
Strange one, now i've imported it i'm getting an issue where outlines are no longer displaying, wonder if you have the same issue @thibaudcolas ?

apps/frontend/static_src/scss/main.scss
```diff
@@ -4,6 +4,7 @@
 @import "./variables";
 @import "./functions";
 @import "./mixins";
+@import "./utilities";
```
thibaudcolas, line 7: I’ll move this to be at the end of the file, we want utilities to be able to trump other code.

webpack.config.js
```diff
@@ -48,7 +49,7 @@ const config = {
               sourceMap: true,
               postcssOptions: {
                 plugins: [
-                  "postcss-preset-env"
+                  postcssPresetEnv({ enableClientSidePolyfills: false })
```
thibaudcolas, line 52: Added this because of the `:focus-visible` polyfill getting in the way with its `.focus-visible` class: https://github.com/csstools/postcss-plugins/tree/main/plugin-packs/postcss-preset-env#plugins-that-need-client-library

thibaudcolas (approved):

Issue #114, opened 2022-10-04
Adopt Prettier (JS/SCS/MD formatting)


lb-:
It would be good to adopt Prettier to ensure there is a consistent coding style and doing this earlier in the project will avoid larger diffs.

We cannot adopt for HTML but could adopt for Markdown, JavaScript and SCSS files.

**See adoption in Wagtail**

https://github.com/wagtail/wagtail/blob/eccd84ba870bfe679ca8e780a4d3eb1f149c3cb6/package.json#L128

https://github.com/wagtail/wagtail/blob/eccd84ba870bfe679ca8e780a4d3eb1f149c3cb6/package.json#L131

https://github.com/wagtail/wagtail/blob/main/.prettierignore

Issue #151, opened 2022-10-19
Front-end tooling upgrades


thibaudcolas:
Things I think we could do with:

- [x] Prettier setup
- [x] ESLint
- [x] Stylelint
- [ ] nvm -> fnm https://github.com/wagtail/guide/pull/185

---

De-scoped / future improvements split out

- [ ] #111
- [ ] #188
- [ ] #189

lb-:
Thanks. I'd already created individual issues so I'll work on them bit by bit. 

I'll also move us to `npm` unless there is a huge pushback

lb-:
Some PRs up - will let these get merged or get feedback and look at the other parts later.

* https://github.com/wagtail/guide/pull/159
* https://github.com/wagtail/guide/pull/156
* https://github.com/wagtail/guide/pull/159

thibaudcolas:
🙌 I hadn’t realised there were existing issues

thibaudcolas:
@lb- what do you think of splitting the CI build between frontend and backend checks once you revisit yarn/npm? Right now it looks like we’re running front-end linting tasks last even though they’re one of the fastest checks.

Here is a sample split setup in GitHub Actions if it helps: https://github.com/torchbox/storybook-django/blob/main/.github/workflows/ci.yml

lb-:
@thibaudcolas  good call, we also did a similar thing for the Hallo package
https://github.com/wagtail/wagtail-hallo/blob/main/.github/workflows/ci.yml

lb-:
https://github.com/wagtail/guide/pull/185 added for `fnm` usage. Once this is merged we can close this issue as the other items are a bit more nice to have and probably not blockers for release.

PR #157, opened 2022-10-19
Adopt Prettier (code formatting)

lb-:
* Resolves https://github.com/wagtail/guide/issues/114
* Relates to https://github.com/wagtail/guide/issues/151


package.json
```diff
@@ -1,46 +1,50 @@
...
+  "private": true,
+  "author": "",
+  "engines": {
+    "node": "^16"
+  },
+  "scripts": {
+    "start": "webpack --mode=development --watch --progress",
+    "build": "webpack --mode=production",
+    "format": "prettier --write .",
+    "lint:format": "prettier --check .",
```
thibaudcolas, line 12: Note I think this won’t process dotfiles (`.gitpod.yml`). If we wanted this we’d need a more complex pattern like the one we use in Wagtail.

README.md
```diff
@@ -3,31 +3,32 @@
...
 
-- Python >= 3.9
-- Poetry
-- Git
-- Node 16.*
-- [Yarn](https://yarnpkg.com/)
+-   Python >= 3.9
+-   Poetry
+-   Git
+-   Node 16.\*
```
thibaudcolas, line 30: Oh interesting

thibaudcolas (approved):
Looking great! One thing I think we could consider revisit, let me know what you think @lb-.

README.md
lb-: yeah true, I did not even look at that - not sure why Prettier changed this.

package.json
lb-: ahh ok - I always wondered why you did that.

Have raised https://github.com/wagtail/guide/issues/162 for an improvement.

PR #158, opened 2022-10-19
Add Poetry version to readme

lb-:
> This is a single commit built built on a different PR https://github.com/wagtail/guide/pull/157

* First time using Poetry for me and spent about 30 minutes on various errors to find that Poetry 1.2.2 is required due to some backwards incompatibility issues.
* I have built on https://github.com/wagtail/guide/pull/157 with a small change to the Readme that just clarifies the version to use.

## Some notes / links

https://github.com/python-poetry/poetry/issues/4983

```
poetry install
...setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(

  RuntimeError

  The Poetry configuration is invalid:
    - Additional properties are not allowed ('group' was unexpected)
  

  at ~/.poetry/lib/poetry/_vendor/py3.10/poetry/core/factory.py:43 in create_poetry
       39│             message = ""
       40│             for error in check_result["errors"]:
       41│                 message += "  - {}\n".format(error)
       42│ 
    →  43│             raise RuntimeError("The Poetry configuration is invalid:\n" + message)
       44│ 
       45│         # Load package
       46│         name = local_config["name"]
       47│         version = local_config["version"]
make: *** [backend] Error 1

```


thibaudcolas (approved):
Sorry about that LB! Agree this is an important point as our configuration will definitely fail with v1.1 and earlier.

lb-:
Curious @thibaudcolas  - why are we using Poetry for this Wagtail project when we do not use it for any others (maybe we do, but I have not used it in the Wagtail repos I have worked on).

thibaudcolas:
@lb- two reasons:

- For a site, we need locking of all transitive dependencies, so pip alone wouldn’t cut it – we’d have to set up either pip-tools, pipenv, a bespoke `pip freeze` workflow – or Poetry.
- The wagtail.org rebuild uses Poetry (https://github.com/wagtail/wagtail.org/tree/next)

lb-:
Ahh I see thanks Thibaud

PR #161, opened 2022-10-20
Feature/search overlay

nicklee:
Updating the search overlay:

 - Removing bs classes
 - Hiding button if js is enabled
 - Moving search count inside input to make it obvious it's an autocomplete
 - Increasing response time of autocomplete
 - Removing aria label on search page (as text is sufficient), only non-js users will see this page so not much time has been spent styling it

### DIFF
--- a/apps/frontend/static_src/scss/components/autocomplete.scss
+++ b/apps/frontend/static_src/scss/components/autocomplete.scss
@@ -4,6 +4,10 @@
 
     &__count {
         @include fs(s);
+        position: absolute;
+        top: $gutter;
+        right: 0;
+        color: $color--grey;
         font-weight: 500;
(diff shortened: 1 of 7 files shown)


apps/frontend/static_src/scss/components/autocomplete.scss
```diff
@@ -0,0 +1,58 @@
+.autocomplete {
+    $root: &;
+    $row-spacing: ($gutter * 2);
+
+    &__count {
+        @include fs(s);
+        position: absolute;
+        top: $gutter;
+        right: 0;
```
thibaudcolas:
```suggestion
        inset-inline-end: 0;
```

thibaudcolas (approved):

PR #163, opened 2022-10-20
Feature/fe launch checklist

nicklee:
- Adds title and meta descriptions
- Add social media share info and image


apps/core/templates/base.html
```diff
@@ -30,6 +30,25 @@
...
+        <meta name="twitter:title" content="{{ page.title }}" />
+        <meta name="twitter:description" content="{% if page.search_description %}{{ page.search_description }}{% else %}{{ page.listing_summary }}{% endif %}">
+        <meta name="twitter:image" content="{{ current_site.root_url }}{% static 'images/social-image.png' %}">
+
+        {# facebook opengraph tags #}
+        <meta property="fb:app_id" content="{{ settings.utils.SocialMediaSettings.facebook_app_id }}" />
+        <meta property="og:type" content="website" />
+        <meta property="og:url" content="{{ site.root_url }}{{ page.url }}" />
+        <meta property="og:title" content="{{ page.title }}" />
+         <meta name="og:image" content="{{ current_site.root_url }}{% static 'images/social-image.png' %}">
```
lb-, line 48:
```suggestion
        <meta name="og:image" content="{{ current_site.root_url }}{% static 'images/social-image.png' %}">
```

Just an extra space here I think

apps/core/templates/base.html
```diff
@@ -30,6 +30,25 @@
...
         {% endif %}
 
+        {% comment %}
+            Twitter summary card - see https://dev.twitter.com/cards/types/summary
+            and https://dev.twitter.com/cards/getting-started
+        {% endcomment %}
+        <meta name="twitter:card" content="summary" />
+        <meta name="twitter:site" content="@wagtailcms" />
+        <meta name="twitter:title" content="{{ page.title }}" />
+        <meta name="twitter:description" content="{% if page.search_description %}{{ page.search_description }}{% else %}{{ page.listing_summary }}{% endif %}">
```
thibaudcolas, line 40: Same as above, we don’t have a `listing_summary` on this build.

apps/core/templates/base.html
```diff
@@ -30,6 +30,25 @@
...
+        <meta name="twitter:image" content="{{ current_site.root_url }}{% static 'images/social-image.png' %}">
+
+        {# facebook opengraph tags #}
+        <meta property="fb:app_id" content="{{ settings.utils.SocialMediaSettings.facebook_app_id }}" />
```
thibaudcolas, line 44: We don’t have this set up on this site, and likely never will.

```suggestion
```

thibaudcolas (changes requested):
I think there’s a bit more to do here, as we don’t have some of those fields. Also the favicon setup seems partial?

apps/core/templates/base.html
```diff
@@ -30,6 +32,24 @@
...
 
+        {% comment %}
+            Twitter summary card - see https://dev.twitter.com/cards/types/summary
+            and https://dev.twitter.com/cards/getting-started
+        {% endcomment %}
+        <meta name="twitter:card" content="summary" />
+        <meta name="twitter:site" content="@wagtailcms" />
+        <meta name="twitter:title" content="{{ page.title }}" />
+        <meta name="twitter:description" content="{% if page.search_description %}{{ page.search_description }}{% endif %}">
+        <meta name="twitter:image" content="{{ current_site.root_url }}{% static 'images/social-image.png' %}">
```
thibaudcolas, line 43: I doubt this would work as our Webpack tooling doesn’t have support for images. I’ll move the file to the `public` folder later.

thibaudcolas (approved):

apps/core/templates/base.html
thibaudcolas: My bad, this works, just needs to use `manifest` for correctness. It wasn’t working in Gitpod for other reasons (site record)

PR #159, opened 2022-10-19
Adoption of Eslint for code quality checks

lb-:
- Install & Configure Eslint - to align with other Wagtail projects using https://github.com/wagtail/eslint-config-wagtail
- Update variable & function names to align with Eslint
- Update cookies function to avoid using loop and instead use functional style
- Reorder main.js functions to avoid using hoisting / globals
- Remove unused variables/imports
- Adds some JSDOC syntax for better code documentation
- Fixes #112
- Relates to #151
- Builds on PR https://github.com/wagtail/guide/pull/157 & https://github.com/wagtail/guide/pull/158

PR #165, opened 2022-10-20
Add Cloudflare worker

RealOrangeOne:
This adds a Cloudflare caching worker from https://github.com/torchbox/cloudflare-recipes/blob/master/common-caching.js

Note: Deployment is a manual process, requiring the Torchbox sysadmin team.


thibaudcolas (approved):
Thanks @RealOrangeOne! I just merged JS linting checks in #159 that might clash with this once merged on `main`. We’ll see what happens.

Issue #113, opened 2022-10-04
Adopt Stylelint (SCSS/CSS linting) with shared Wagtail linting rules


lb-:
It would be good to adopt Stylelint to ensure there is a consistent coding style but also to avoid pitfalls with CSS that could cause accessibility or localisation issues.

* https://github.com/wagtail/stylelint-config-wagtail

PR #171, opened 2022-10-20
Adoption of Stylelint for SCSS/CSS quality checks

lb-:
* Fixes https://github.com/wagtail/guide/issues/113
* Some files had to be ignored as I was not sure what do to with existing styles, not too critical though.
* Some `important` usage appears to be required so I just ignored a few discrete lines
* Fixed up a few random typos I found in the styles files also, hope that is ok

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,37 @@
+{
+  "extends": ["@wagtail/stylelint-config-wagtail"],
+  "rules": {
+    "scale-unlimited/declaration-strict-value": [
+      ["/color/", "fill", "stroke"],
+      {
+        "ignoreValues": [
+          "currentColor",
+          "inherit",
+          "initial",
+          "none",
+          "unset",
+          "transparent",
+          "Canvas",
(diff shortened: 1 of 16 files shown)


lb-:
Note the lint will fail due to other JS changes - will see if I can fix those up in a separate PR

.stylelintrc.json
```diff
@@ -0,0 +1,37 @@
+{
+  "extends": ["@wagtail/stylelint-config-wagtail"],
+  "rules": {
+    "scale-unlimited/declaration-strict-value": [
+      ["/color/", "fill", "stroke"],
+      {
+        "ignoreValues": [
```
thibaudcolas, line 7: I’m surprised you chose JSON for this file! Not a blocker but it’d be nice to be able to comment on why specific rules are overridden like this.

thibaudcolas (approved):

.stylelintrc.json
lb-: Yeah I just went for the simplest way to get the config going

PR #173, opened 2022-10-20
Rework caching-worker to align with Eslint usage

lb-:
* Adopts linting for JS file introduced in https://github.com/wagtail/guide/pull/165
* Note: We could also just add this file to the eslintignore if it is not meant to align with our linting styleguide


thibaudcolas:
@lb- I will close this now and ignore the file instead. Linting this file will likely introduce more problems than is worthwhile – the execution environment of Cloudflare workers is very different to browsers (I believe they use V8 isolates with standard library additions). So we’d need quite a bit of configuration for the linting to even produce useful feedback – which doesn’t feel worthwhile.

As an example – there’s no `window` object there, so `addEventListener` is correct.

lb-:
Thanks @thibaudcolas  agreed.

PR #175, opened 2022-10-21
Extract translatable strings from templates

### LINKED ISSUE #146: Words for translation
A list of the hardcoded language throughout the site:

### Header

- Wagtail user guide
- User guide
- Toggle the search form
- Search
- Quick links
- Skip to main content

### Feedback widget

- Was this page helpful?
- Help us improve by telling what you missed
- Submit
- happy
- unhappy

### Footer

- About this guide
- Learn why we structured our documents like this
- Help improve this guide
- Contribute to open source with no coding experience needed
- Visit Wagtail.org for more resources and Wagtail news

### Homepage

- This documentation is written for anyone creating content or managing content production in Wagtail. It covers everything you can expect when using a standard Wagtail site.

### Content page

- On this page

laymonage:
Fixes #146.

TODO:
- Use machine translation for initial `.po` files


saevarom (approved):
I think this looks good, apart from the failing CI

apps/core/migrations/0006_footertext.py
```diff
@@ -0,0 +1,49 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+import uuid
+import wagtail.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("wagtailcore", "0077_alter_revision_user"),
```
allcaps, line 12: Depending om `wagtailcore.0077` gives me: `ValueError: Related model 'wagtailcore.pagerevision' cannot be resolved` when running the tests.

Not sure why we need this dep.

apps/core/models/__init__.py
```diff
@@ -1,7 +1,11 @@
...
+from .feedback import Feedback
 from .home import HomePage
+from .snippets import FooterText
 
 __all__ = [
-    HomePage,
     ContentPage,
+    Feedback,
+    FooterText,
+    HomePage,
```
allcaps, line 10: These need to be strings. Se also: https://github.com/wagtail/guide/pull/184

apps/custom_media/locale/af/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,22 @@
+# SOME DESCRIPTIVE TITLE.
```
allcaps, line 1: I'm not sure why you provide all locales. It generates a lot of work whenever we update a translation string. Why not only support languages that will have translated content?

locale/pt_pt/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,185 @@
+# SOME DESCRIPTIVE TITLE.
```
allcaps, line 1: Why `pt_pt` and not `pt`?

apps/guide/settings/base.py
```diff
@@ -267,6 +267,10 @@
     for version in WAGTAIL_GUIDE_VERSIONS
 ]
 
+LOCALE_PATHS = [
+    BASE_DIR / "locale",
+]
```
allcaps, line 272: Why do we have `custom_media/locale/`? Can't everything go in the global locale folder? Would be nice to have a single po per language.

apps/core/templates/500.html
```diff
@@ -1,13 +1,14 @@
...
         <meta name="viewport" content="width=device-width, initial-scale=1" />
     </head>
     <body>
-        <h1>Internal server error</h1>
+        <h1>{% trans 'Internal server error' %}</h1>
 
-        <h2>Sorry, there seems to be an error. Please try again soon.</h2>
+        <h2>{% trans 'Sorry, there seems to be an error. Please try again soon.' %}</h2>
     </body>
 </html>
```
allcaps, line 14: I'm a bit hesitant to put any 'logic' in 500.html. The server crashed. It might be better to not depend on anything.

I would keep this English.

allcaps (changes requested):
Thank you for the work!

Checking this out locally, gave me `ValueError: Related model 'wagtailcore.pagerevision' cannot be resolved` when running tests. Not sure why, removing a migration dep solved the issue. See my comment.

I propose to not translate the `500.html` and keep it plain html.

Can we add `makemessages` and `compilemessages` to the Makefile? For easy development?

Don't put too much effort in auto translations. I'd prefer to see this merged. We can always auto translate po's later, and/or translate by hand (minor effort compared to content translation).

Not sure why we have locale folder in custom_media.

apps/core/migrations/0006_footertext.py
laymonage: Yeah that was autogenerated with `makemigrations`, I think it should be safe to remove. It does highlight an issue with wagtail-localize though (https://github.com/wagtail/wagtail-localize/issues/642), which otherwise I wouldn't have found.

However I'm thinking of removing this entirely in favour of #177.

apps/custom_media/locale/af/LC_MESSAGES/django.po
laymonage: I generated them based on our `WAGTAIL_CONTENT_LANGUAGES`, but if we want to keep this to a minimum until someone translates the content to the other languages, I'm fine with that too

locale/pt_pt/LC_MESSAGES/django.po
laymonage: No real reason, just that Wagtail itself uses `pt-pt` for Portuguese (Portugal): https://github.com/wagtail/wagtail/blob/7286c530e97e2eb64f3747c6904a7456789f82b7/wagtail/admin/localization.py#L34

and that we've already configured the settings to use `pt-pt`:

https://github.com/wagtail/guide/blob/8689a4ec7aacc8bef1034c6a91e3cad9a30ff24c/apps/guide/settings/base.py#L242

Django does use `pt` though:

https://github.com/django/django/blob/main/django/conf/locale/__init__.py#L455


P.s. `pt-pt` still works because `get_language_info()` falls back to `pt`

allcaps: I'd always go for single language codes as `pt-pt` doesn't provide a fallback for `pt-br` and `pt` does. Same for any other language/region combination where the language is the same as the region.

FYI: Wagtail had `nl_nl` and `nl` and it was very confusing with nasty side effects.

On top of that, shorter URLs are nicer. ;)

apps/core/migrations/0006_footertext.py
allcaps: I had to debug for quite a while to find this dep was the issue. 

Remove the footer logic, or not, I'll leave up to you! FWIW, your work seems more complete then #177

apps/custom_media/locale/af/LC_MESSAGES/django.po
allcaps: Let's go with what you have now.

We can also drop languages in the future, if it is clear that they won't be used and maintaining empty languages becomes annoying.

apps/core/migrations/0006_footertext.py
laymonage: Oh I mean just the footer model, not the PR itself. The snippet models in #177 also covers the three footer items as well, while mine only covers the copyright text.

apps/core/templates/500.html
thibaudcolas: Seems fair to me, but then let’s add a comment to this file to explain why it’s hard-coded to english.

apps/custom_media/locale/af/LC_MESSAGES/django.po
thibaudcolas: For this app in particular – I don’t think we need the `_("image")` verbose_name to be translate-able. Is that even displayed anywhere in the UI? This feels like a lot of files created for the single low-value string.

apps/core/templates/components/footer.html
```diff
@@ -38,14 +39,14 @@ <h2 class="link-block__heading">Help improve this guide</h2>
                     <div class="link-block__content">
                         <h2 class="link-block__heading">Wagtail</h2>
                         <div class="link-block__description">
-                            Visit Wagtail.org for more resources and Wagtail news
+                            {% trans 'Visit Wagtail.org for more resources and Wagtail news' %}
```
thibaudcolas, line 42: Here and above – I’d have expected all of this to be done with a Footer snippet.

laymonage: This should be covered in #177, I'll rebase this PR once that's merged.

locale/pt_pt/LC_MESSAGES/django.po
laymonage: Okay, I've renamed our config to use `pt` for Portuguese (Portugal).

apps/guide/settings/base.py
laymonage: Looks like it's because `custom_media/locale/` already exists. I've deleted it so that all locale strings are generated to `BASE_DIR / "locale"` only.

apps/custom_media/locale/af/LC_MESSAGES/django.po
allcaps: > When several applications provide different versions of the same resource (template, static file, management command, translation), the application listed first in [INSTALLED_APPS](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-INSTALLED_APPS) has precedence.

Via: https://docs.djangoproject.com/en/4.1/ref/settings/#installed-apps

So you can leave the string marked for translation, as Wagtail PO's will translate this string.
We only need to delete to remove the locale folder.

thibaudcolas (approved):

PR #177, opened 2022-10-23
make footer editable

Hitansh-Shah:
Hey @thibaudcolas. I have registered a snippet `FooterContent` to allow footer to be editable. As you suggested, I have created a template tag `get_footer_content` which renders the footer. I was facing some issues with streamfield and structblock inside snippet so I went ahead with Oderables. The remaining task is to figure out how to handle icons in the footers. I tried to include a field for choosing image but for some reason it was rendering as a dropdown/select field with no entries in admin. I also found a similar [issue](https://github.com/wagtail/wagtail/issues/1912) being raised some time ago. Therefore for now I have just used same image for all the icons. Let me know for any improvements.


Hitansh-Shah:
Conflicts ☹️. I guess I forgot to take the latest pull ;). I will resolve them soon.

apps/core/templatetags/core_tags.py
```diff
@@ -25,3 +25,14 @@ def navigation(context):
...
+
+@register.inclusion_tag("components/footer.html", takes_context=True)
+def get_footer_content(context):
+    footer = ""
+    if FooterContent.objects.first() is not None:
+        footer = FooterContent.objects.first()
+
+    return {
+        "footer": footer,
+    }
```
laymonage, line 38: Drive-by comment: we could improve this by reducing the duplicated query. And it seems in the template, we don't really handle the case where the `FooterContent` object isn't available, so we might as well skip the `if`...?

```suggestion
    return {
        "footer": FooterContent.objects.first(),
    }
```

Hitansh-Shah: Thanks @laymonage. I will update the PR

Hitansh-Shah:
@thibaudcolas @laymonage I have updated the PR. Please have a look

apps/core/migrations/0006_footercontent_footeritem.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.0.7 on 2022-10-23 06:39
+
+from django.db import migrations, models
+import django.db.models.deletion
+import modelcluster.fields
+import wagtail.fields
+
+
+class Migration(migrations.Migration):
```
laymonage, line 9: Could you please add a separate migration that creates the current content as an instance of this snippet? If not I can add it later when I merge it

apps/core/models/snippets.py
```diff
@@ -0,0 +1,35 @@
...
+from modelcluster.fields import ParentalKey
+from modelcluster.models import ClusterableModel
+from wagtail.admin.panels import FieldPanel, InlinePanel
+from wagtail.fields import RichTextField
+from wagtail.models import Orderable
+from wagtail.snippets.models import register_snippet
+
+
+@register_snippet
+class FooterContent(ClusterableModel):
```
laymonage, line 12: I think it'd be nice to have a `__str__` method here that just returns `"Footer content"`, so that it doesn't say `FooterContent object (1)` in the admin

apps/core/templates/components/footer.html
```diff
@@ -1,51 +1,26 @@
+{% load wagtailcore_tags %}
 
 <div class="footer">
     <div class="footer__container">
         <div class="footer__link-blocks">
+          {% for footer_item in footer.footer_items.all %}
             <div class="link-block">
```
laymonage, line 7: I don't see the `target_link` being used, can we turn this into an `<a>` element with the `target_link` as the `href`?

apps/core/templates/components/footer.html
```diff
@@ -1,51 +1,26 @@
...
                 <div class="link-block__container">
                     <div class="link-block__icon">
                         {% include "./svg/info.svg" %}
```
laymonage, line 10: All of the items now use the same icon, is this intended? If not we could add a `CharField` to the snippet model for the icon and set the `choices` to a predefined `TextChoices` class:

https://docs.djangoproject.com/en/4.1/ref/models/fields/#enumeration-types

apps/core/templatetags/core_tags.py
```diff
@@ -27,6 +27,12 @@ def hreflang_url(value, arg):
     return value.replace(f"/{arg}/", "/en-latest/", 1)
 
 
+@register.inclusion_tag("components/footer.html", takes_context=True)
+def get_footer_content(context):
```
laymonage, line 31: Maybe rename it to just `footer` so it matches with `navigation`?

laymonage (commented):
Works pretty well, great work! Some minor comments about the links and icons for the `FooterItem`.

apps/core/models/snippets.py
laymonage, line 12: Oh and before I forget: please also make this class and `FooterItem` extend `wagtail.models.TranslatableMixin`, and regenerate the migrations so we can make sure that they are translatable

apps/core/migrations/0006_footercontent_footeritem.py
Hitansh-Shah: Hey @laymonage . I could not exactly understand what you meant here so I have left this task 😅 . Could you please elaborate on what should I do here? Or if you want you can pick this up.

Hitansh-Shah:
Hey @laymonage. I have completed all the tasks except for the "separate migration for current content ...". Please have a look.

apps/core/migrations/0006_footercontent_footeritem.py
laymonage: We can write our own migration and use `migrations.RunPython` to create an initial instance of the snippet based on the content that we have now. For reference: https://github.com/wagtail/guide/pull/175/commits/7bb24aabd440284532e26a064574bf6144ed4d31#diff-18e703351254cbef277b981e328c2a4941262dd85442cdd93c980255b76d67ba

But no worries, I'm happy with the PR overall. I'll write the migration along with some minor fixes to the styling of links in the footer text 😄

laymonage (approved):
Looks good to me! I'll make some minor fixes and write a migration to create an initial instance of the snippet.

laymonage:
@Hitansh-Shah Thanks! FYI I renamed the fields in `FooterItem` to be more concise, so to have these changes on your local machine, I'd suggest running `python manage.py migrate core 0005_alter_feedback_options` (to rollback to the migration before the models were created) and `python manage.py migrate` again.

Hitansh-Shah:
@laymonage Gotcha 🙌 .

Issue #189, opened 2022-10-24
Split the CI build between frontend and backend checks


lb-:
Once https://github.com/wagtail/guide/issues/111 is completed we should look at splitting the CI build between frontend and backend checks.

Right now it looks like we’re running front-end linting tasks last even though they’re one of the fastest checks.

We should also ensure we are using the `nvmrc` as the node version reference - see `node-version-file: '.nvmrc'` & https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

Example implementations

* https://github.com/torchbox/storybook-django/blob/main/.github/workflows/ci.yml
* https://github.com/wagtail/wagtail-hallo/blob/main/.github/workflows/ci.yml

_Originally posted by @thibaudcolas in https://github.com/wagtail/guide/issues/151#issuecomment-1285157031_

PR #191, opened 2022-10-24
Rename test.yml to ci.yml

lb-:
- File contains more than testing and this aligns with other Wagtail projects naming of the same kind of file
- Prep for #189 (best to have this as a separate commit so we do not lose the file history)


allcaps (approved):

lb-:
conflicts fixed - merging in - thanks @allcaps

PR #193, opened 2022-10-25
Optimize static images

zerolab:
Screenshot 2022-10-25 at 10 22 42

Saved 5 out of 24KB. While not massive, it adds up over time


thibaudcolas (approved):
Nice one @zerolab!

PR #196, opened 2022-10-25
Split CI steps

### LINKED ISSUE #189: Split the CI build between frontend and backend checks

lb-:
- Fixes #189
- Adds titles to each job
- Split out the Python (Django) tests to their own job
- Split backend/frontend linting to their own isolated jobs
- Allow make format & make lint to be run as frontend/backend only versions also

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,16 +12,24 @@ test:
 test-coverage:
 	poetry run coverage run manage.py test && poetry run coverage report
 
-format:
+format-backend:
 	poetry run isort apps
 	poetry run black apps
+
+format-frontend:
 	npm run format
 
-lint:
+format: format-backend format-frontend
+
+lint-backend:
 	poetry run flake8 apps
 	poetry run isort --check-only --diff apps
 	poetry run black --check --diff apps
+
+lint-frontend:
 	npm run lint
+
+lint: lint-backend lint-frontend
 
 frontend:
 	npm ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   test:
+    name: 📋 Test
     runs-on: ubuntu-latest
 
     env:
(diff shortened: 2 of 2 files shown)


lb-:
OK this should be good to review now 🤞

Makefile
```diff
@@ -15,13 +15,21 @@ test-coverage:
...
 
 lint:
 	poetry run flake8 apps
 	poetry run isort --check-only --diff apps
 	poetry run black --check --diff apps
+ifeq ($(backend-only),true)
+	echo "skipping frontend linting"
+else
 	npm run lint
+endif
```
laymonage, line 32: How about splitting these into `format-backend`, `format-frontend`, `lint-backend`, `lint-frontend`, and then making `format` and `lint` call both the `backend` and `frontend` variants?

Kind of like bakerydemo: https://github.com/wagtail/bakerydemo/blob/main/Makefile

lb-: Yeah we did a similar thing in the Wagtail repo also (split into different make commands), but I recall that being not ideal as it added so many different commands.

lb-: will do this though

laymonage (approved):

allcaps:
TY! :clap:

Issue #179, opened 2022-10-24
Address accessibility feedback from audit


thibaudcolas:
Audit by Scott: #76.

thibaudcolas:
Copying the findings from Scott,

---

## WAVE

### Errors
- [x] Search input field is missing a label
- [x] Feedback form textarea (only on content pages, only visible after clicking frowny face) is missing a label
- [x] `en-latest` is not a valid value for the `lang` attribute on the `html` element
  - Related to #16 and #143. I am not very familiar with Wagtail localization or know if this combo of language and version is a common pattern, but it seems like using an invalid value for `html lang` could be problematic.
- [x] `aria-labelledby="searchModalLabel"` on `<div class="modal fade" id="search-modal" tabindex="-1" aria-labelledby="searchModalLabel" aria-hidden="true">` is invalid (and, I believe, unnecessary)

### Warnings
- There are four screenshots flagged as having "long" alt text, but I feel that this is appropriate for this sort of documentation.


## Firefox dev tools accessibility panel

- [x] SVG elements (e.g., homepage illustrations, note header ℹ️ icons) are presumed to have ARIA role of diagram and thus it's problematic that they have no text label. Should likely be marked as `aria-hidden`.
- [x] Broken links in rich text render an `<a>` element with no `href` -- this is a reasonable approach


## Axe

- [x] There is no `main` landmark on the homepage. There is one on content pages, but it's just an extra wrapper. Should replace `<div class="app__content" id="main-content">` in both page templates with `<main class="app__content" id="main-content">` and remove the extra `main` from the content page template.
- [x] The note header ℹ️ icons have `id="Layer_1"` attributes, which creates duplicate `id`s when more than one are on a page.


## Pa11y

### Tweaks we could make to hit Level AAA
- [x] Update homepage section description text color (e.g., "Our tutorials will help you learn and master the different ways of creating and managing content in Wagtail. It’s the best place to get started.") to `--w-color-grey-600`.


## Manual observations

### Keyboard navigation
- [x] User bar should be moved after the "Skip to main content" link, per the [guidance added](https://docs.wagtail.org/en/stable/topics/writing_templates.html#wagtail-user-bar) added when [Storm made it more accessible in 2.14](https://docs.wagtail.org/en/stable/releases/2.14.html#user-bar-with-keyboard-and-screen-reader-support).
[...]

PR #204, opened 2022-10-25
Feature/179 accessibility audit

nicklee:
Fixing issues raised here:

https://github.com/wagtail/guide/issues/179


nicklee:
Worth noting that the search icon won't display, but there is a fix for this in the dark/light MR (the search svg was being served as an img tag instead of an inline svg)

allcaps (approved):
Approved, I'll leave this MR open so that people who raised the issue can have a look.

thibaudcolas (approved):

PR #206, opened 2022-10-25
Move table of contents generation to `save_revision()`

### LINKED ISSUE #182: Translating table of contents
How will we translate the `table_of_contents` field on `ContentPage`? See for example https://guide.wagtail.org/is-latest/tutorials/getting-started/ where I've translated the available content to Icelandic, but the table of contents remains in English.

I can't see that this has been handled in #175 and I'm not sure why this field does not show up in the `wagtail-localize` interface.

What would be the best strategy to translate this type of data? I'm not sure there is a good way to handle it in Wagtail. 
Perhaps skip translations by dynamically extracting the headlines in Javascript instead of doing it on save?

laymonage:
Fixes #182.

This ensures that table of contents are also automatically translated. Wagtail Localize uses a custom form, so we cannot rely on a custom `WagtailAdminPageForm` to generate the table of contents. By moving it to `save_revision()`, table of contents are generated for every revision of the instance, which is created on every edit (including through Localize).


allcaps:
Beautiful :clap: TY!

Issue #195, opened 2022-10-25
hreflang link meta tags should be created for all locales of a given page


thibaudcolas:
Right now we only do this for the english page: https://github.com/wagtail/guide/blob/21ea3ae7bdbfb787101a7c38886db801a5a6ad8a/apps/core/templates/base.html#L30-L33

Instead, we should create a `link` tag with `hreflang` for all available versions of the page, just like how we populate the language selector.

See also #194.

PR #208, opened 2022-10-25
Fix html `lang` attribute and generate `hreflang` links to translations

### LINKED ISSUE #194: Use the correct locale code in html lang attribute and language switcher hreflang
Spotted as part of #125.

- [ ] The `<html>` lang attribute shouldn’t contain the version suffix (currently `en-latest`)
- [ ] The `hreflang` attribute should be populated with locale codes (currently empty)

laymonage:
Fixes #194 and #195.

Needs tests.


thibaudcolas (approved):
Looking excellent @laymonage!

PR #217, opened 2022-10-26
Use original id when generating links in table of contents

### LINKED ISSUE #182: Translating table of contents

laymonage:
Fixes #182.

I spent some time digging into this, looks like wagtail-localize completely skips over custom rich text features we added here:

https://github.com/wagtail/guide/blob/60cb0bb278e1a295b295734acb51d0e0ba7ded00/apps/richtext/wagtail_hooks.py#L7-L41

I haven't found a way to fix this properly, as it seems that it's just how wagtail-localize works, i.e. it separates out the HTML stuff from user-facing strings.

This PR adds a workaround by using the existing id (in the original language) so that links will still work (though the anchor slug will be in the original language, e.g. English).

This issue only happens with synced translations. Unsynced translations work normally.

~~I'll try to add tests tomorrow.~~ Added.

### DIFF
--- a/apps/core/models/content.py
+++ b/apps/core/models/content.py
@@ -4,7 +4,7 @@ from bs4 import BeautifulSoup
 from django.db import models
 from django.http import HttpResponse
 from django.template import Context, Template
-from django.template.defaultfilters import slugify
+from django.utils.text import slugify
 from wagtail.admin.panels import FieldPanel
 from wagtail.core.fields import StreamField
 from wagtail.models import Page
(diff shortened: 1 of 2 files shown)


saevarom (approved):
I've tested this locally and the links work in this PR. Looks good to me.

PR #218, opened 2022-10-26
Search only in the active locale

### LINKED ISSUE #209: Localize search results, only search current locale
Pages from all locales are returned. Search for 'Wagtail' and you get English and Icelandic results. We want only results of the current locale.

> To implement this, open up search/views.py in your favourite editor, then modify the following lines:

> Change the import line from wagtail.core.models import Page to from wagtail.core.models import Page, Locale
> Change the query Page.objects.live().search(search_query) to Page.objects.live().filter(locale=Locale.get_active()).search(search_query)

Via https://www.wagtail-localize.org/tutorial/4-templates/ (bottom of the page)

saevarom:
Fixes #209

### DIFF
--- a/apps/core/templates/base.html
+++ b/apps/core/templates/base.html
@@ -81,6 +81,7 @@
         {% include "components/search_modal.html" %}
 
         {# Global javascript #}
+        <script>var languageCode = '{{ self.locale.language_code }}';</script>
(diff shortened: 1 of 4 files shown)


thibaudcolas (approved):
All working as expected, thank you @saevarom

PR #221, opened 2022-10-27
Add support for docker-compose. 

saevarom:
An iteration on some Makefile commands I added earlier. I realised that it was too hard to manage the container using only docker, it's much nicer to use docker-compose. This makes it easier to manage the container with volumes to make the database persistent, and for developers to make changes to the editor guide code and test them out on the running container.

I did have to make my own version of the Dockerfile but only because I couldn't override the `POETRY_INSTALL_ARGS` setting in the Dockerfile.

### DIFF
--- a/Dockerfile.localdev
+++ b/Dockerfile.localdev
@@ -1,0 +1,98 @@
+# (Keep the version in sync with the node install below)
+FROM node:16 as frontend
+
+# Make build & post-install scripts behave as if we were in a CI environment (e.g. for logging verbosity purposes).
+ARG CI=true
+
+# Install front-end dependencies.
+COPY package.json package-lock.json webpack.config.js ./
+RUN npm ci
+
+# Compile static files
+COPY ./apps/frontend/static_src/ ./apps/frontend/static_src/
+RUN npm run build
+
+
+# We use Debian images because they are considered more stable than the alpine
+# ones becase they use a different C compiler. Debian images also come with
+# all useful packages required for image manipulation out of the box. They
+# however weigh a lot, approx. up to 1.5GiB per built image.
+FROM python:3.9 as production
+
+ARG POETRY_HOME=/opt/poetry
+ARG POETRY_INSTALL_ARGS=""
+
+# IMPORTANT: Remember to review this when upgrading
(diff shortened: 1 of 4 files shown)


Dockerfile.localdev
```diff
@@ -0,0 +1,98 @@
+# (Keep the version in sync with the node install below)
```
RealOrangeOne, line 1: Question: What's the value in having a separate dockerfile? The main one ought to work just fine for this.

Makefile
```diff
@@ -45,13 +45,21 @@ run:
...
-	docker exec -it guide_latest /bin/bash
+docker-shell:
+	docker-compose exec web bash
+
+docker-shell-frontend:
+	docker-compose exec frontend bash
 
 docker-init:
-	docker exec -it guide_latest /bin/bash -c "make backend"
+	docker-compose exec web poetry install
```
RealOrangeOne, line 60: Question: Why are dependencies being run like this? It happens during build time, which is probably fine.

```suggestion
```

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+volumes:
+  node_modules:
+services:
+  web:
+    build:
+      context: .
+      dockerfile: Dockerfile.localdev
+    extra_hosts:
+      - 'host.docker.internal:host-gateway'
+    command: poetry run python manage.py runserver 0.0.0.0:${PORT:-8000}
```
RealOrangeOne, line 12: Issue: If we always override the command here, why not just put it in the dockerfile?

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    volumes:
+      - .:/app:delegated,rw
+      - node_modules:/app/node_modules/
+    ports:
+      - '${PORT:-8000}:${PORT:-8000}'
+    environment:
+      SECRET_KEY: $SECRET_KEY
```
RealOrangeOne, line 19: Why are we duplicating these to pass them through? I'd recommend either using dummy values, or having `docker-compose` read from a `.env` file.

thibaudcolas:
Superseded with #433

PR #223, opened 2022-10-28
Fix home page crash when translation cannot be found

laymonage:
Fixes https://sentry.io/organizations/wagtail-org/issues/3699443518/?project=4504043711037440

Honestly don't know why this happens as Django's `get_language()` function should fall back to `settings.LANGUAGE_CODE`, and I thought we could assume the `HomePage` with the default `LANGUAGE_CODE` (i.e. `en-latest`) exists. Not sure how to test it either as this doesn't seem to happen locally. I'll add another PR if I find a way to test this.


laymonage:
Merging this immediately as it's a fairly straightforward fix so that the site isn't down, will investigate more later.

PR #224, opened 2022-10-28
Override default richtext fullwidth rendition to 900px

### LINKED ISSUE #216: Image renditions use the wrong width
We seem to be rendering all images in the content with a `width-800` rendition (the default in Wagtail for full-width rich text images). This is problematic because we display images at a width of 900px at most, so on desktop all images look blurry.

We should switch to a `width-900` rendition instead.

laymonage:
Fixes #216.


thibaudcolas (approved):
Thank you @laymonage! This looks so, so much better.

PR #232, opened 2022-10-28
Adding a missing fieldpanel

saevarom:
The introduction was moved to a field but it is missing a FieldPanel for editing


thibaudcolas (approved):
😅

PR #239, opened 2022-10-29
Fix docker setup for wagtail guide

VictoriaAjala:
This PR fixes issues I encountered when setting up wagtail with docker.

Without installing the necessary dependency with 

```
RUN apt-get install -y gettext 
```

the `make docker-init` step fails with the error below:
Screenshot from 2022-10-30 00-16-28

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN npm run build
 FROM python:3.9 as production
 
 # Install dependency needed to run guide init script
-RUN apt-get install -y gettext 
+RUN apt-get update && apt-get install -y gettext  
 
 ARG POETRY_HOME=/opt/poetry
 ARG POETRY_INSTALL_ARGS="--without dev"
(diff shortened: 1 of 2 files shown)


VictoriaAjala:
@lb- @thibaudcolas

lb-:
I do not use docker, so I am not the best to review this - but I am sure someone in the team will see this and give feedback when available.

saevarom:
I'd prefer not messing with the Heroku Dockerfile since that is used for deployments. 
There is another PR #221 which defines a separate Dockerfile for local development and there we could add suport for gettext in the container.

So I would recommend merging #221 first and then think about supporting gettext in the local docker container. It would probably make the most sense to close this PR and create a new one after #221 is merged.

VictoriaAjala:
Alright @saevarom, sounds good.

Closing this and creating a new PR when #221 has been merged.

Issue #243, opened 2022-10-31
Footer in mobile view should be in column


VictoriaAjala:
present
![Screenshot from 2022-10-31 14-47-20](https://user-images.githubusercontent.com/102407780/199023096-ec0b798a-e7e7-412b-92f8-fe76bb0f361f.png)

proposed
![Screenshot from 2022-10-31 15-57-27](https://user-images.githubusercontent.com/102407780/199040152-12f324d3-3e86-4a57-94fd-09124d2971e5.png)

PR #244, opened 2022-10-31
Changed footer view to column in mobile view

VictoriaAjala:
Before
Screenshot from 2022-10-31 15-59-28

After
Screenshot from 2022-10-31 15-57-27
Fixes issue #243


VictoriaAjala:
@Scotchester okay

PR #254, opened 2022-11-04
Translate apps only

allcaps:
This PR makes the `make translations` command cd into the apps folder to translate only our strings.
It also drops the file locations (comments with line numbers) from the po files to avoid git conflicts.


saevarom (approved):
This looks solid to me. Great work!

PR #256, opened 2022-11-05
Fix dropdown item hover bgcolor in dark mode

allcaps:
# Before

Screenshot 2022-11-05 at 02 08 19

# After

Screenshot 2022-11-05 at 02 08 40

### DIFF
--- a/apps/frontend/static_src/scss/components/language-selector.scss
+++ b/apps/frontend/static_src/scss/components/language-selector.scss
@@ -49,5 +49,11 @@
             background-color: $color--black;
             color: $color--light-grey;
         }
+        #{$root}__dropdown-item {
+            &:hover,
+            &:focus {
+                background-color: $color--grey;
+            }
+        }
     }
 }


apps/frontend/static_src/scss/components/language-selector.scss
```diff
@@ -49,5 +49,11 @@
             background-color: $color--black;
             color: $color--light-grey;
         }
```
nicklee, line 51: Small nitpick: suprised the linter didn't suggest having a space between these two?

nicklee:
LGTM, nice one!

Scotchester:
How did the alignment of the dropdown change from aligned on the right side of the button to aligned on the left side of the button? 🤔

allcaps:
> How did the alignment of the dropdown change from aligned on the right side of the button to aligned on the left side of the button? 🤔

I don't know, must be something during the screenshot, as the Gitpod and the live site behave the same.

PR #257, opened 2022-11-05
Add middleware to check for published homepage or 404

### LINKED ISSUE #253: Language redirection active even though language not translated
Currently when visiting https://guide.wagtail.org/ it redirects to the browser preferred language and shows that language in the dropdown even though that language is not available yet. Which is a bit confusing.
E.g. preferred browser language German redirects to https://guide.wagtail.org/de-latest/ even though the content is not ready yet and displays "Deutsch (latest)" in the language drop down. See screenshot below.

Also, English is not available in the language selection list.

It would be good to disable languages that are not yet translated to avoid confusion I think?

image

allcaps:
Fixes #253

### DIFF
--- a/apps/core/middleware.py
+++ b/apps/core/middleware.py
@@ -1,0 +1,14 @@
+from django.shortcuts import get_object_or_404
+from django.utils.translation import get_language
+
+from apps.core.models import HomePage
+
+
+class ValidateLocaleMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        get_object_or_404(HomePage, locale__language_code=get_language(), live=True)
+        response = self.get_response(request)
+        return response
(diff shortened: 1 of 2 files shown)


apps/core/middleware.py
```diff
@@ -0,0 +1,17 @@
...
+from apps.core.models import HomePage
+
+
+class ValidateLocaleMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if lang_code := get_language_from_path(request.path_info):
+            print(lang_code)
```
allcaps, line 13: Err, print

apps/core/middleware.py
```diff
@@ -0,0 +1,17 @@
...
+
+from apps.core.models import HomePage
+
+
+class ValidateLocaleMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if lang_code := get_language_from_path(request.path_info):
```
allcaps, line 12: Since we are after `LocalMiddleware, we can use `django.utils.translation.get_language()` to get the current language.

allcaps:
@saevarom Thank you for the review! I've updated this PR and switched to the `get_language()` method.

saevarom (approved):
Tested again, looks good

PR #260, opened 2022-11-09
Add section page

allcaps:
I think we should start thinking about the exact design/contents of SectionPages.

Maybe:
- TOC always off
- Introduction text introducing the sub pages.
- Call to action blocks to link to the subpages.

Not sure, let us discuss!


thibaudcolas:
I assume this draft is out of date so will close now

PR #264, opened 2022-11-10
Remove language selector when we don't have a page to show versions of.

### LINKED ISSUE #263: Language picker inactive when user is redirected to 404 page
**Steps to reproduce**
1. Set your browser language to german
2. Enter https://guide.wagtail.org/
3. As German is probably already configured as available language but the German index is not yet published, I am redirected to https://guide.wagtail.org/de-latest/ which triggers a 404

**Expected result**
I can access the guide in another language by selecting i.e. English in the language picker.

**Actual result**
Other languages are not available in the Language Picker

saevarom:
Fixes #263 

I think it is pointless to show a button with only the active locale (which might not have any pages at all) and simpler to just remove in those cases. The only case that I know of right now is the 404.html template which is just a dumb template that has no idea of where to direct the user to go.

I added a bit more logic in the homepage selection middleware introduced in #257 so that the english language is activated instead of the incorrectly selected one. That way the user has the site tree to the left to navigate to the english version of the guide.

### Before
Screenshot 2022-11-10 at 10 36 09

### After
Screenshot 2022-11-10 at 10 37 18

### DIFF
--- a/apps/core/middleware.py
+++ b/apps/core/middleware.py
@@ -12,7 +12,7 @@ class ValidateLocaleMiddleware:
         try:
             HomePage.objects.get(locale__language_code=get_language(), live=True)
         except HomePage.DoesNotExist:
-            activate('en-latest')
+            activate("en-latest")
             raise Http404()
         response = self.get_response(request)
         return response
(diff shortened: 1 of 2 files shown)


allcaps (commented):
LGTM, merged!

PR #267, opened 2022-11-11
Fixes invalid redirection for /admin when browser is not set to English

### LINKED ISSUE #262: Invalid redirect for Wagtail Admin when browser language is not English
**Steps to reproduce**
1. Set your browser language to German
2. Open https://guide.wagtail.org/admin 

**Expected result**
Login view for Wagtail Admin is shown

**Actual Result**
Request is redirected to https://guide.wagtail.org/de-latest/admin/ which results in a 404

_Workaround: Set browser language to English to acccess the admin_

saevarom:
Fixes #262.

Move the homepage lookup after response handling so that we have access to the url resolver. Then check if the requested path belongs to i18n_patterns so that we are not handling e.g. `/admin` as a localized URL.

To test this you will need to:

1. Make sure the browser language is english
    1.  Test that `/admin` works properly
    2. Test that  `/` works properly and redirects to `/en-latest``
2. Change browser language to Deutsch
    1. Test that `/admin` works properly 
    2. Test that `/` redirects to `/de-latest/` and shows a 404 page with the English menu

### DIFF
--- a/apps/core/middleware.py
+++ b/apps/core/middleware.py
@@ -14,10 +14,11 @@ class ValidateLocaleMiddleware:
     def __call__(self, request):
         response = self.get_response(request)
         if request.resolver_match:
-            # Check if the path is in the i18n_patterns 
+            # Check if the path is in the i18n_patterns
             if pattern.match(request.resolver_match.route):
                 try:
-                    HomePage.objects.get(locale__language_code=get_language(), live=True)
+                    HomePage.objects.get(
+                        locale__language_code=get_language(), live=True)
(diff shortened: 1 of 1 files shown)


saevarom:
> Great! Thank you! Is it okay if I write some tests for this?

Sure

thibaudcolas (approved):
👍 this seems to be working great.

thibaudcolas:
I see the tests are still in progress in #277, but the fix itself seems worthwhile to include already, so I’ll merge now.

PR #277, opened 2022-11-18
Fix 262 tests

allcaps:


allcaps:
@saevarom Can you have a look at this one?

apps/core/middleware.py
```diff
@@ -1,18 +1,29 @@
...
+        ):
+            try:
+                HomePage.objects.get(locale__language_code=get_language(), live=True)
+                response = self.get_response(request)
+            except HomePage.DoesNotExist:
+                # The requested language is not available, use the default
+                activate(settings.LANGUAGE_CODE)
+                response = self.get_response(request)
+        else:
+            response = self.get_response(request)
```
saevarom, line 28: If you call the resolver before `get_response` you will always have None, so this code defaults to the else statement every single time. I ran the tests with a print statement in the else section and it printed every time.

I also tried to set my language to a language that had a matching homepage and it did not properly set my preferred language until I put the `response = self.get_response(request)` at the top.

I see that you removed the 404 response when a language does not exist, was there a reason for that?

saevarom: I'm not sure why the tests do not fail on the language preference, but at least it was not working in the browser for me

saevarom: I think that the tests that correctly pass when a user asks for "/" and is redirected to the preferred language (browser or cookie) because of LocaleMiddleware

saevarom: So, I guess we need to take a step back and actually think about what the requirements are

saevarom (changes requested):
See comments in middleware

thibaudcolas:
pretty sure this is out of date 👁️

PR #310, opened 2023-06-01
redirect to default language instead of a 404 error

disperate:
If the root URL is opened with an HTTP_ACCEPT_LANGUAGE where no translation is available, a 404 error is returned. 
See issue https://github.com/wagtail/guide/issues/308.

Instead of a 404 error, users are now redirected to the default language.


laymonage (approved):
This works well, tested it and it does fix the issue. Thank you very much for adding the test as well!

PR #375, opened 2023-11-03
Custom 404

Tijani-Dia:
The goal of this PR is to display relevant translations of a page when it's not available in a specific locale.

In the example below, the `Search` page isn't available in Dutch for Wagtail 5.1:
127 0 0 1_8000_nl-5 1 x_tutorial_search_


thibaudcolas (approved):
Thank you @Tijani-Dia!

Issue #409, opened 2025-11-19
Change homepage introduction to rich text


thibaudcolas:
We use the website to occasionally promote things to our audience. For example Wagtail Space, or getting feedback on specific aspects of the project.

It would be nice if we didn’t have to hard-code homepage content for this. Update the existing `introduction` field to support rich text:

https://github.com/wagtail/guide/blob/8b0f9d00d57aa7159b731f8df06ab7e73a0a4654/apps/core/models/home.py#L15

As part of implementing this, it would be nice if there was a text that the field’s rich text content does render as expected.

PR #411, opened 2025-11-20
Fix "Change homepage introduction to rich text"

eigenwijsje:
This 

- Changes the `HomePage.introduction` field into a `wagtail.model.fields.RichTextField`
- Updates the template to render the introduction as safe HTML and inside a `<div>`,
- Adds a test that the homepage with introduction renders

Fixes: https://github.com/wagtail/guide/issues/409


thibaudcolas (approved):
All working well! A few small changes needed

PR #428, opened 2026-04-07
docs: update Poetry env activation command for Poetry 2.x

Raghaddahi:
### Description

update Poetry env activation command for Poetry 2.x as the poetry shell command is no longer used.

Since Poetry 2.0.0, the `poetry shell` command is no longer installed by default.

This replaces it with `source $(poetry env activate)`, which is the recommended approach.

Documentation: https://python-poetry.org/docs/managing-environments/#activating-the-environment

### AI usage
None

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ You can access the administrative area at [http://127.0.0.1:8000/admin](http://1
 
 To activate Poetry's virtual environment, run:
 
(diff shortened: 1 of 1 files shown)


Raghaddahi:
Hi @thibaudcolas, could you review this when you get a chance? Thanks!

PR #453, opened 2026-07-01
Remove page-based versioning system, add URL bridge middleware (#452)

### LINKED ISSUE #452: Remove current versioning system from the site
Part of #451. Follow-up to #445. We want to remove all code of the existing page-based versioning system. This will help us figure out what steps to take once we do so in production, so we can remove the code and content without breaking the site for people currently navigating to a specific version.

For example, if I navigate to https://guide.wagtail.org/en-6.0.x/how-to-guides/manage-snippets/, I should arrive on https://guide.wagtail.org/en/how-to-guides/manage-snippets/?target_version=6.0.x

Once we have that in place, we won’t need:

- Code that uses `WAGTAIL_GUIDE_VERSIONS`
- monkey patches
- ???
- buildfixtures logic to create versioned content

## Working on this

Assigning to @Raghaddahi

Raghaddahi:
Fixes #452

## Description

Removes the existing page-based versioning system (which treated Wagtail versions as fake Locales via `LANGUAGE_CODE` suffixes like `en-6.0.x`) and replaces it with a lightweight URL bridge middleware. Old version-suffixed URLs (e.g. `/en-6.0.x/how-to-guides/manage-snippets/`) now 302-redirect to the bare-locale URL with a `?target_version=` query param: `/en/how-to-guides/manage-snippets/?target_version=6.0.x`.

### Why this approach

The fake-locale design required monkey-patching Django's language-code regexes, a cartesian product of `LANGUAGES × WAGTAIL_GUIDE_VERSIONS`, duplicate HomePage/content trees per version, and complex version-combo fallback logic. Collapsing to bare locales (`en`, `nl`, `pt-br`) eliminates all of that while the bridge middleware keeps existing external `/en-X.x/` links working — staged rollout without breaking saved URLs.

### Areas requiring careful review

- **`apps/core/middleware.py`** — new `VersionedUrlRedirectMiddleware`. Regex is strict (`latest | N.N.x | N.x`), non-GET passes through (no form-submit redirects), 302 (to be flipped to 301 after a week of clean logs). Wired ABOVE `LocaleMiddleware` so the dotted prefix never reaches locale activation.
- **`apps/guide/settings/base.py`** — `WAGTAIL_GUIDE_VERSIONS` deleted, `LANGUAGES = WAGTAIL_GUIDE_LANGUAGES` (bare codes), `LANGUAGE_CODE = "en"`. This is the root unblock; every downstream edit depends on it.
- **`apps/core/migrations/0007_initial_footercontent.py`** — edited historical migration: hardcoded `"en-latest"` lookup → `"en"`. Safe because prod already has `0007` applied (won't re-run); only fresh test/new installs re-run the edited version.
- **`apps/core/models/home.py`** — `get_fallback_pages` simplified to plain translation fallback (no version matrix). The old `rsplit("-", 1)` would crash on bare locale codes.

### New middleware test file

`apps/core/tests/test_versioned_url_redirect.py` — 7 dedicated tests for `VersionedUrlRedirectMiddleware`:

```bash
poetry run python manage.py test apps.core.tests.test_versioned_url_redirect -v 2
```

### Manual QA worth doing

- `curl -I localhost:8000/en-6.0.x/how-to-guides/manage-snippets/` — confirm 302 to `/en/how-to-guides/manage-snippets/?target_version=6.0.x`.
- Verify admin "View live" links still work (the `LocaleURLMixin` removal affects `ContentPage`'s `get_url_parts`).

### Out of scope (follow-up)

- Production data migration (step 3) — deleting version-suffixed `Locale` rows + page trees from the prod DB. To be done after this is deployed and stable.
- `prompts/content/` regeneration — the `fetch_content.py` script is already updated to use `en`, but re-running it against the live site should happen after deploy. A separate PR will regenerate `prompts/content/en/` and delete the stale `en-latest/` directory.

## AI usage
The
AI (opencode / glm-5.2) assisted with this PR description ,discussion and review throughout the implementation. 
all code is reviewed by me


Raghaddahi:
> Tested the versioning-removal branch against a restored copy of production data. The VersionedUrlRedirectMiddleware correctly redirects old versioned URLs (e.g. /en-6.0.x/...) to the new format (/en/...?target_version=6.0.x), confirmed via curl. However, the destination fails with a 500 Internal Server Error, because production data only contains versioned locales (en-latest, en-4.1.x … en-7.4.x, etc.) — there is no plain en locale for Wagtail to resolve the page against. This branch needs a data migration to consolidate locales (e.g., promote en-latest → en) before it can be deployed against real production data.

claude code

apps/core/middleware.py
```diff
@@ -7,6 +10,31 @@
...
+
+    def __call__(self, request):
+        path = request.path_info
+        match = PATTERN.match(path)
+
+        if match and request.method in ("GET", "HEAD"):
+            lang = match.group("lang")
+            ver = match.group("ver")
+            rest = match.group("rest")
+            parsed = urlparse(rest)
```
thibaudcolas: just `rest` should be enough no?

thibaudcolas (approved):

PR #462, opened 2026-07-10
fix:fallback theme styles for browsers without light-dark() support

### LINKED ISSUE #426: Add fallback theme styles for browsers without light-dark() support
### Issue summary

I was exploring the guide repo’s theming implementation. We currently rely on the modern light-dark() color function for theme rendering, which is great for supported browsers.

In supported browsers (tested in Chrome), this works correctly. In older browesrs (e.g. Edge) versions, theme colors render incorrectly, and dark mode is not displayed as expected.

**I’ve attached screen-video showing:**

Correct behavior in a supported browser `(CSS.supports(...) returns true)`
Incorrect rendering in older Edge where support is missing  `(CSS.supports(...) returns false)
`

https://github.com/user-attachments/assets/6bcb447c-2aa6-4f3b-8de5-6cae9750af72

**Why this matters**
Most users are on updated browsers, but there are still compatibility gaps in older versions. Without fallback styles, parts of the UI can appear broken.

**Proposed solution**
Add a fallback block for unsupported browsers, for example:

`@supports not (color: light-dark(#000, #fff)) { ... }`

Use baseline static light/dark token values in that block.

I tested this fallback approach locally, and it fixes the incorrect rendering in older Edge.

### Steps to reproduce

test it with any browser which doesn't support the `light-dark() `CSS function 
source: https://caniuse.com/mdn-css_types_color_light-dark

### Additional information

_No response_

### Can be reproduced

Not confirmed

### Technical details

- Browser: Chrome, Edge
- Browser version:  Edge 95 and 111

### Working on this

**Needs confirmation**
Should we add a fallback block for older browser versions?

As-ajay06:
Fixes #426

### Description

Add fallback theme styles for browsers without light-dark() support

### AI usage

AI has been for drafting in this issue , and researching the browser part.

### Also regarding new changes to the issue
 
the CSS light-dark() function is a relatively new feature and achieved baseline support across major modern browsers around May 2024.

Here is the breakdown of which older versions will NOT run this feature (i.e. they do not support light-dark()):

+ Chrome & Edge: Versions 122 and older will NOT support it. (Support was introduced in version 123, released in March 2024).
+ Firefox: Versions 119 and older will NOT support it. (Support was introduced in version 120, released in November 2023).
+ Safari: Versions 17.3 and older will NOT support it. (Support was introduced in version 17.4, released in March 2024).
+ Opera: Versions 108 and older will NOT support it. (Because Opera is based on Chromium, it gained support around the same time as Chrome 123, which maps to Opera 109).

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,6 +1,12 @@
 {
   "extends": ["@wagtail/stylelint-config-wagtail"],
   "rules": {
+    "selector-pseudo-class-no-unknown": [
+      true,
+      {
+        "ignorePseudoClasses": ["global"]
+      }
+    ],
     "scale-unlimited/declaration-strict-value": [
       ["color", "/-color/", "fill", "stroke"],
       {
--- a/apps/frontend/static_src/scss/mixins.scss
+++ b/apps/frontend/static_src/scss/mixins.scss
@@ -87,3 +87,18 @@
         }
     }
 }
+
+@mixin light-dark($property, $light, $dark){
+
+    #{$property}: $light;
+    @supports (color: light-dark(#000, #fff)) {
+        #{$property}: light-dark(#{$light}, #{$dark});
+    }
+
+    @supports not (color: light-dark(#000, #fff)) {
+        :global(body.theme-dark) & {
+            #{$property}: $dark;
+        }
+    }
+}
+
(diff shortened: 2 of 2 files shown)


thibaudcolas:
Closing, see #426

Issue #445, opened 2026-06-10
Versioned blocks prototyping


thibaudcolas:
### Problem statement

Part of GSoC project - versioning friction

### Proposed solution

Prototype those two options in one PR:

* **VersionNoteBlock**  
  * Option 1: new specific block type specifically for this, with two fields:  
    * Version number (select dropdown)  
      * Type of change (added/changed/removed)  
      * Content (rich text)  
    * Option 2: new generic block type with support for versioning  
      * Same fields but using [block settings / BlockGroup](https://docs.wagtail.org/en/stable/reference/streamfield/blocks.html#blockgroup)  
  * **wagtail\_version** field for images  
    * Interesting to do this to help with content management but want to find a way that doesn’t require complex integration  
    * Next step: Thibaud refinement requirement for image versioning

### Working on this

Assigning to @Raghaddahi

thibaudcolas:
Notes from #449:

- UI / design: keep the smaller more subtle style, with inline text. The "alert block" style is too distracting. Designing for 4-5 "version added" block highlights per page.
- Backend: make all new block fields optional. So you can more easily toggle a block to have or not have a version highlight
- Report on blocks usage (see [StreamField block type usage report](https://github.com/wagtail/wagtail/issues/11506)). "Which pages have versioned blocks that I should consider updating / removing"

Nice to have / future:

- UI filter to hide content when wanting to look at docs for an older release
- Content design: we need more guidance on how and when to create "version changed" highlights, and how and when to remove them
- Warning message to reduce the risk of people looking at very outdated docs / comparing "latest" docs for a very old version

thibaudcolas:
Now merged! Follow-up in #451

PR #470, opened 2026-08-14
Add Annotated blocks usage report for outdated version content

Raghaddahi:
Relates to #445

### Description

Adds a new **Annotated blocks usage** report to the Wagtail admin (under Reports) listing ContentPages with `text_annotated` StreamField blocks. Each block's version is compared against `WAGTAIL_GUIDE_CURRENT_VERSION` (a new setting) and flagged as **Needs update** when outdated, so editors can find stale version-specific content without manual searching.

Each block links directly to its position in the page editor via `#block-index-N`, with a small JS helper (`block_usage_focus.js`, registered via `insert_editor_js`) that expands and scrolls to the exact block. Block position is used instead of UUID since UUIDs differ between the live page and the latest draft revision.

A second JS/CSS pair (`block_usage_report.js`/`.css`) powers a hover-sync effect: hovering any of the Change type / Version / Content cells highlights all three together.

### Testing
image

image

### AI usage

AI (GLM 5.2) generated the code and PR description, reviewed and assisted throughout by me.

### DIFF
--- a/apps/core/reports.py
+++ b/apps/core/reports.py
@@ -1,0 +1,114 @@
+from django.conf import settings
+from django.forms import Media
+from django.utils.html import strip_tags
+from django.utils.translation import gettext_lazy as _
+from wagtail.admin.views.reports import PageReportView
+from wagtail.permissions import page_permission_policy
+
+from apps.core.models import ContentPage
+
+
+def iter_annotated_blocks(page):
+    """Yield (index, StreamChild) for each text_annotated block in the page's body."""
+    for block_index, block in enumerate(page.body):
+        if block.block_type == "text_annotated":
+            yield block_index, block
+
+
+def _truncate(text, max_length):
+    if len(text) > max_length:
+        return text[:max_length] + "…"
+    return text
+
+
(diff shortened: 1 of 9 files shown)


thibaudcolas (changes requested):
This is nice but I think it’s too much code / too advanced. We should strive to create reports with Python code only, and if crucial for functionality we can introduce HTML too, but adding CSS/JS for reports feels unwarranted. If there is missing functionality to reports, we should add it to Wagtail core. But first, simplify the interface.

For the functionality, I don’t think the "outdated" check is quite right. It is completely normal to highlight major changes in "old" versions, lots of people will still be using those and appreciate the heads'up. Reporting versioned blocks is helpful, but flagging some of this versioning usage as outdated doesn’t feel right to me. I’d prefer we altogether remove that.

Note this is different from what I had in mind in #445 with a warning message. The warning would be for _visitors to the site_ that use an old Wagtail version, to let them know that version might no longer be supported. That could be worthwhile but I don’t think this is high enough value to warrant building.

